PR #118, opened 2021-06-21
Fix web assets path

Robert-Steiner:
`flutter build web` and `flutter run -d chrome` behave differently when bundling the web assets. In `flutter run -d chrome` we can access an asset via `http://localhost:56864/assets/assets/qambert_v0001/qambert.onnx` or `http://localhost:56864/assets/qambert_v0001/qambert.onnx`. With `flutter build web`, on the other hand, we can only access an asset via `http://localhost:56864/assets/assets/qambert_v0001/qambert.onnx`. 

There are several tickets that relate to this problem:
- https://github.com/flutter/flutter/issues/80784
- https://github.com/flutter/flutter/issues/67655

In this PR we change the `_baseAssetUrl` from `/assets` to `/assets/assets` so that it works for both commands.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -93,7 +93,7 @@ dependencies = [ "build-dart" ]
 script_runner = "@duckscript"
 env = { WASM_SUFFIX="wasm_bindings" }
 script = ["""
-  wasm_rel_path = set "bindings/dart/example/${WASM_SUFFIX}"
+  wasm_rel_path = set "bindings/dart/example/assets/${WASM_SUFFIX}"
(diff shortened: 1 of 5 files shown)


Robert-Steiner:
I just rebased and rechecked each target. Looks good🚀

longtimedeveloper:
@Robert-Steiner do you know when this fix will be in the stable branch?  

I'm on the current stable and have the same problems with my images.  It's good that we can manually adjust the build output to work around the issue.

acrrd:
> @Robert-Steiner do you know when this fix will be in the stable branch?

It's better if you ask on some flutter issue related to this problem.

PR #134, opened 2021-06-28
TY-1877 exclude dart example in release repo

Robert-Steiner:
**Ticket:**

[TY-1877]

**Summary**

- exclude dart example 
- exclude building [wasm android library](https://github.com/xaynetwork/xayn_ai_release/tree/staging/android/src/main/jniLibs/arm64-v8a)
  - if we were to create a new release now, the wasm lib would still appear as it is in the build cache (the `target` folder)
  - however i will change the cache key in [TY-1871] so no need to do it in this pr

[TY-1877]: https://xainag.atlassian.net/browse/TY-1877

[TY-1871]: https://xainag.atlassian.net/browse/TY-1871

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -615,7 +615,7 @@ jobs:
           # --ignore-unmatch avoid to fail if the repository is empty
           git rm --ignore-unmatch -r .
 
-          rsync -a --exclude example ${{ env.DART_WORKSPACE}}/ .
+          rsync -a --exclude example ${{ env.DART_WORKSPACE }}/ .
 
           # Remove files from .gitignore that needs to be uploaded to the release repo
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -615,7 +615,7 @@ jobs:
           # --ignore-unmatch avoid to fail if the repository is empty
           git rm --ignore-unmatch -r .
 
-          rsync -a ${{ env.DART_WORKSPACE}}/ .
+          rsync -a --exclude example ${{ env.DART_WORKSPACE}}/ .
```
janpetschexain, line 618:
```suggestion
          rsync -a --exclude example ${{ env.DART_WORKSPACE }}/ .
```

janpetschexain (approved):

PR #136, opened 2021-06-30
TY-1871 use github.job in cache key

Robert-Steiner:
**Ticket**

[TY-1871]

**Summary**
- use `github.job` instead of a hardcoded string to make it less error-prone when copying an exisiting job to use it as a template for a new job 

[TY-1871]: https://xainag.atlassian.net/browse/TY-1871

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -38,10 +38,10 @@ jobs:
           all_but_latest: true
           access_token: ${{ github.token }}
 
-  registry-cache:
+  cargo-registry-cache:
     # we use the latest stable rustc + cargo version that is already installed on the image
     # https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#rust-tools
-    name: cargo-fetch
+    name: cargo-registry-cache
     runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -333,9 +333,9 @@ jobs:
           if-no-files-found: error
           path: target/${{ matrix.target}}/release/libxayn_ai_ffi_c_${{ matrix.target }}.a
 
-  build-wasm-lib-release:
-    name: build-wasm-lib-release
-    needs: registry-cache
+  build-release-wasm-lib:
```
janpetschexain, line 336: should we already use the plural here (and across all ci files)? we probably have to support `wasm64` next to the current `wasm32` in the near future anyways, then the ci naming could stay the same. this would also align the naming with the other platform libs job names.
```suggestion
  build-release-wasm-libs:
```

janpetschexain (approved):

.github/workflows/release.yaml
Robert-Steiner: I think using wasm64 is only required if we need more than 4 gb of ram, but I can change the name to be consistent with the other jobs

PR #137, opened 2021-06-30
Run tests for all rerank mode

acrrd:
Run tests on the reranker with both `Search` and `News` mode.

### DIFF
--- a/xayn-ai/src/lib.rs
+++ b/xayn-ai/src/lib.rs
@@ -43,6 +43,7 @@ mod tests;
 // which is not possible to silence.
 #[cfg(test)]
 #[allow(unused_imports)]
+#[rustfmt::skip]
 pub(crate) use rstest_reuse as rstest_reuse;
 
 // Reexport for assert_approx_eq for usage in FFI crates.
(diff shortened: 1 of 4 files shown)


xayn-ai/src/lib.rs
```diff
@@ -38,6 +38,13 @@ pub use crate::{
 #[cfg(test)]
 mod tests;
 
+// we need to export rstest_reuse from the root for it to works.
+// `use rstest_reuse` will trigger clippy::single_component_path_imports
+// that is not possible to silence.
```
janpetschexain, line 43:
```suggestion
// we need to export rstest_reuse from the root for it to work.
// `use rstest_reuse` will trigger `clippy::single_component_path_imports`
// which is not possible to silence.
```

xayn-ai/src/lib.rs
```diff
@@ -38,6 +38,13 @@ pub use crate::{
...
+#[cfg(test)]
+#[allow(unused_imports)]
+pub(crate) use rstest_reuse as rstest_reuse;
```
janpetschexain, line 46: since `rustfmt` wants to remove same name aliases, but we need the alias for the aforementioned reasons, our only option seems to silence the formatting as well.
```suggestion
#[rustfmt::skip]
pub(crate) use rstest_reuse as rstest_reuse;
```

janpetschexain (approved):

PR #138, opened 2021-06-30
TY 1928 With QA-mBert use title if snippet is empty.

acrrd:
For some type of result (video, images, ...) we don't have snippets but only the title.
When running QA-mBert we use the title if the snippet is empty.

### DIFF
--- a/xayn-ai/src/embedding/qambert.rs
+++ b/xayn-ai/src/embedding/qambert.rs
@@ -18,7 +18,15 @@ impl QAMBertSystem for QAMBert {
             documents
                 .into_iter()
                 .map(|document| {
-                    self.run(&document.document_content.snippet)
+                    let snippet = &document.document_content.snippet;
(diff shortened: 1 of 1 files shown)


xayn-ai/src/embedding/qambert.rs
```diff
@@ -18,7 +18,15 @@ impl QAMBertSystem for QAMBert {
             documents
                 .into_iter()
                 .map(|document| {
-                    self.run(&document.document_content.snippet)
+                    let snippet = &document.document_content.snippet;
+                    // not all documents have a snippet, if snippet is empty we use the title
+                    let data = if !snippet.is_empty() {
```
Robert-Steiner, line 23: Maybe a little easier to read but i’m fine either way
```rust
let data = if snippet.is_empty() {
    &document.document_content.title
} else {
    snippet
};
```

Robert-Steiner (approved):

PR #139, opened 2021-06-30
TY 1927 Change weights to  compute contex value

acrrd:
We want the similarity from the QA-mBert to weight 70% of the context value, all the other component equally share the rest.


rustonaut (approved):
lgtm

So just to be on the safe side is following correct?

- 70% qambert
- 20% smbert/COI (10% pos, 10% neg)
- 10% domain reranker

acrrd:
> So just to be on the safe side is following correct?
> 
>     * 70% qambert
> 
>     * 20% smbert/COI (10% pos, 10% neg)
> 
>     * 10% domain reranker

Yes

PR #142, opened 2021-07-06
Ty 1923 list net training pr 1 (back-propagation forward & back-ward pass)

rustonaut:
Implements training on `ListNet`.

Including:

- forward pass
- backward pass
- evaluation
- a simple "training" function

This does not include a API on the Domain Reranker or feature extraction
as this parts depend a lot on the XaynNet and Team Blue integration.


xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -494,4 +970,301 @@ mod tests {
...
+        fn begin_of_training(&mut self, _list_net: &ListNet) {
+            dbg!("begin of training");
+        }
+
+        fn end_of_training(&mut self, _list_net: &ListNet) {
+            dbg!("end of training");
+        }
+    }
+
+    //FIXME create better tests
```
janpetschexain, line 1084: it would be nice to have a set of tests which directly compares the result of a single forward and a backward pass for one input batch against the expected result.

xayn-ai/src/ltr/list_net/ndlayers/dense.rs
```diff
@@ -110,15 +124,116 @@ where
...
+        let mut out = input.dot(&self.weights);
+        out += &self.bias;
+        if for_back_propagation {
+            z_out = Some(out.clone());
+        }
+        let y_out = self.activation_function.apply_to(out);
+        (y_out, z_out)
+    }
+
+    /// This calculates the gradients of a dense layer based on the relevant partial derivatives.
```
Robert-Steiner, line 152:
```suggestion
    /// Calculates the gradients of a dense layer based on the relevant partial derivatives.
```

xayn-ai/src/ltr/list_net/ndutils/mod.rs
```diff
@@ -28,6 +28,30 @@ where
...
 
+/// Compute the Kullback-Leibler Divergence between a "good" distribution and one we want to evaluate.
+///
+/// This returns a result based on `nats`, i.e. it uses `ln` (instead of `log2` which
+/// would produce a result based on `bits`).
+///
+/// All values are clamped/cliped to the range `f32::EPSILON..=1.`.
+///
+/// For the eval distribution this makes sense as we should never predict `0` but at most
+/// a value so close to it, that it ends up as 0 due to the limited precision of
```
Robert-Steiner, line 39:
```suggestion
/// a value so close to it, that it ends up as `0` due to the limited precision of
```

xayn-ai/src/ltr/list_net/ndutils/mod.rs
```diff
@@ -28,6 +28,30 @@ where
...
+/// `f32`.
+///
+/// For the good distribution we could argue similarly. An alternative choice
+/// is to return 0 if the good distributions probability is 0.)
```
Robert-Steiner, line 43:
```suggestion
/// is to return `0` if the good distributions probability is `0`.
```

xayn-ai/src/ltr/list_net/ndutils/mod.rs
```diff
@@ -28,6 +28,30 @@ where
     array
 }
 
+/// Compute the Kullback-Leibler Divergence between a "good" distribution and one we want to evaluate.
+///
+/// This returns a result based on `nats`, i.e. it uses `ln` (instead of `log2` which
```
Robert-Steiner, line 33:
```suggestion
/// Returns a result based on `nats`, i.e. it uses `ln` (instead of `log2` which
```

xayn-ai/src/ltr/list_net/ndutils/mod.rs
```diff
@@ -28,6 +28,30 @@ where
     array
 }
 
+/// Compute the Kullback-Leibler Divergence between a "good" distribution and one we want to evaluate.
```
Robert-Steiner, line 31:
```suggestion
/// Computes the Kullback-Leibler Divergence between a "good" distribution and one we want to evaluate.
```

xayn-ai/src/ltr/list_net/ndlayers/dense.rs
```diff
@@ -110,15 +124,116 @@ where
...
+        let partials = partials.into_shape((1, weights_nr_columns)).unwrap();
+        let weight_gradients = input.dot(&partials);
+
+        DenseGradientSet {
+            weight_gradients,
+            bias_gradients,
+        }
+    }
+
+    /// Add given gradients to the weight and bias matrices.
```
Robert-Steiner, line 188:
```suggestion
    /// Adds given gradients to the weight and bias matrices.
```

xayn-ai/src/ltr/list_net/ndlayers/dense.rs
```diff
@@ -110,15 +124,116 @@ where
         }
     }
 
-    pub(crate) fn run<S, D>(&self, input: ArrayBase<S, D>) -> Array<f32, D>
+    /// Return the result of applying this dense layer on given inputs.
```
Robert-Steiner, line 127:
```suggestion
    /// Returns the result of applying this dense layer on given inputs.
```

xayn-ai/src/ltr/list_net/ndlayers/activation.rs
```diff
@@ -21,6 +22,20 @@ where
     }
 }
 
+impl Relu {
+    /// Calculates the partial derivatives of reLU at given input.
+    ///
+    /// I.e. returns an array where for all values in the input an 1 is included
```
Robert-Steiner, line 28:
```suggestion
    /// I.e. it returns an array where for all values in the input an 1 is included
```

xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -227,14 +381,335 @@ pub enum LoadingListNetFailed {
...
+#[cfg_attr(test, derive(Debug, Clone))]
+pub struct GradientSet {
+    dense1: DenseGradientSet,
+    dense2: DenseGradientSet,
+    scores: DenseGradientSet,
+    prob_dist: DenseGradientSet,
+}
+
+impl GradientSet {
+    /// Merge all gradient sets computed for one batch of training data.
```
Robert-Steiner, line 417:
```suggestion
    /// Merges all gradient sets computed for one batch of training data.
```

xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -227,14 +381,335 @@ pub enum LoadingListNetFailed {
...
+    optimizer: O,
+}
+
+impl<D, C, O> ListNetTrainer<D, C, O>
+where
+    D: DataSource,
+    C: Callbacks,
+    O: Optimizer,
+{
+    /// Create a new `ListNetTrainer` instance.
```
Robert-Steiner, line 486:
```suggestion
    /// Creates a new `ListNetTrainer` instance.
```

xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -227,14 +381,335 @@ pub enum LoadingListNetFailed {
...
+    /// Resets the "iteration" of training and evaluation samples.
+    ///
+    /// This is allowed to also *change* the training and evaluation
+    /// samples and/or their order. E.g. this could shuffle them
+    /// before every epoch.
+    ///
+    /// This is called at the *begin* of every epoch.
+    fn reset(&mut self);
+
+    /// Return the next `batch_size` number of training samples.
```
Robert-Steiner, line 607:
```suggestion
    /// Returns the next `batch_size` number of training samples.
```

xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -227,14 +381,335 @@ pub enum LoadingListNetFailed {
...
+    ///
+    /// Returns a empty vector once all training samples have been returned.
+    ///
+    /// # Panics
+    ///
+    /// A batch size of 0 is not valid. And implementors are allowed to
+    /// panic if it's passed in.
+    fn next_training_batch(&mut self, batch_size: usize) -> Vec<Sample>;
+
+    /// Return the next evaluation sample.
```
Robert-Steiner, line 617:
```suggestion
    /// Returns the next evaluation sample.
```

xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -227,14 +381,335 @@ pub enum LoadingListNetFailed {
...
+    fn end_of_epoch(&mut self, list_net: &ListNet, mean_kv_divergence_evaluation: Option<f32>);
+
+    /// Called at the begin of training.
+    fn begin_of_training(&mut self, list_net: &ListNet);
+
+    /// Called after training finished.
+    fn end_of_training(&mut self, list_net: &ListNet);
+}
+
+/// Prepare the inputs for training.
```
Robert-Steiner, line 647:
```suggestion
/// Prepares the inputs for training.
```

xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -227,14 +381,335 @@ pub enum LoadingListNetFailed {
...
+///
+/// This is a static helper function which is meant to be used by `DataSource`
+/// implementations to filter and prepare inputs.
+///
+/// This will:
+///
+/// - return `None` if there are less then 10 documents
+/// - truncates inputs and relevances to 10 documents
```
Robert-Steiner, line 655:
```suggestion
/// - truncate inputs and relevances to 10 documents
```

xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -227,14 +381,335 @@ pub enum LoadingListNetFailed {
...
+/// - return `None` if there are no relevant documents (after truncating)
+/// - calculates the target distribution based on the relevances
```
Robert-Steiner, line 657:
```suggestion
/// - calculate the target distribution based on the relevances
```

Robert-Steiner (commented):
Looks good👍
I've added a few suggestions to make the documentation a little more consistent

xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -198,12 +234,130 @@ impl ListNet {
...
+            scores_y,
+            mut prob_dist_y,
+        } = forward_pass;
+
+        let nr_documents = inputs.shape()[0];
+
+        let p_cost_and_prob_dist = {
+            prob_dist_y -= &target_prob_dist;
+            prob_dist_y
+        };
```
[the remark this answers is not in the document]
rustonaut: Got point, it can be simplified again (I started out with what you proposed but at some point this didn't work due to some things related to ndarray trait impl, references and views.).

xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -198,12 +234,130 @@ impl ListNet {
...
+        let p_scores = p_cost_and_prob_dist.dot(&self.prob_dist.weights().t());
+
+        let mut d_scores = DenseGradientSet::zero_gradients_for(&self.scores);
+        let mut d_dense2 = DenseGradientSet::zero_gradients_for(&self.dense2);
+        let mut d_dense1 = DenseGradientSet::zero_gradients_for(&self.dense1);
+
+        for row in 0..nr_documents {
+            // From here on training is "split" into multiple parallel "path" using
+            // shared weights (hence why we add up the gradients).
+            let p_scores = p_scores.slice(s![row..row + 1]);
```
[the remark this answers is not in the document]
rustonaut: Due to the small size of the weight matrix involved on that layer I'm not sure if specializing `gradients_from_partials_0d`  gives any relevant performance improvement. I will consider it.

xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -227,14 +381,335 @@ pub enum LoadingListNetFailed {
...
+    prob_dist: DenseGradientSet,
+}
+
+impl GradientSet {
+    /// Merge all gradient sets computed for one batch of training data.
+    ///
+    /// This will create the mean of each gradient across all gradient sets.
+    ///
+    /// If there are no gradient sets in the input then `None` is returned.
+    fn merge_batch(gradient_sets: impl IntoIterator<Item = GradientSet>) -> Option<GradientSet> {
```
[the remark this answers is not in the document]
rustonaut: Good idea, that might be useful.

We don't really know anything about the range but this kind of neuronal network are known to more likely not work well if the gradients get too far away from 0. If we fully re-train ListNet we should consider introducing batch normalization which could be useful in many ways for that use-case (and fix at some of the more serve problems with the feature extraction :wink: ).

xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -227,14 +381,335 @@ pub enum LoadingListNetFailed {
...
+                },
+            )
+        })
+        .fold((0f32, 0), |(acc, count), cost| (acc + cost, count + 1));
+
+        if count > 0 {
+            Some(acc / count as f32)
+        } else {
+            None
+        }
```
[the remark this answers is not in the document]
rustonaut: I would prefer it keep it this way as "in general" people don't consent over 0 counting as positive or not. (Rusts standard library counts `0` as neither positive nor negative, but then e.g. the `f32` version of `is_positive` is `is_sign_positive` and does count `0` as positive (or negative depending on which f32 zero you have).

xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -227,14 +381,335 @@ pub enum LoadingListNetFailed {
...
+    /// This is allowed to also *change* the training and evaluation
+    /// samples and/or their order. E.g. this could shuffle them
+    /// before every epoch.
+    ///
+    /// This is called at the *begin* of every epoch.
+    fn reset(&mut self);
+
+    /// Return the next `batch_size` number of training samples.
+    ///
+    /// Returns a empty vector once all training samples have been returned.
```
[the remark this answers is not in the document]
rustonaut: I had that before but `Some(vec![])` would be a invalid but possible invariant. 

It also made some of the (not yet opened) follow up PR's code simpler.

xayn-ai/src/ltr/list_net/mod.rs
rustonaut: I have that in another branch and it's problematic.

The problem is that FFN learn to "cope with rounding errors" (oversimplified), but this means that over enough iterations even just the difference of rounding errors between different matmul implementations can make the network diverge to a point where they are comparable (but potentially still yield similar good results).

Even for a single batch with a single sample the results can diverge noticeable between the rust and the tensorflow implementations, reasons include some small hidden implementation differences around tensorflow clipping some values in some place and some other think.

So the test was "flaky"  dependent on the weights used for randomly initializing it, because of which it's currently not included as it at best would be misleading (looking like a test which show that it works but actually only working due to the initial weights chosen).

xayn-ai/src/ltr/list_net/ndlayers/dense.rs
```diff
@@ -110,15 +124,116 @@ where
...
+    ) -> DenseGradientSet {
+        let weights_nr_rows = self.weights.shape()[0];
+        let weights_nr_columns = self.weights.shape()[1];
+        assert_eq!(input.shape()[0], weights_nr_rows);
+        assert_eq!(partials.shape()[0], weights_nr_columns);
+
+        // The formula for bias at index i is:  `b_i = x_i + b_i` of which the derivative wrt. b_i is `1`
+        let bias_gradients = partials.to_owned();
+
+        // For the weight matrix we need the Jacobian Matrix, i.e. donating the inputs as `s_i`
```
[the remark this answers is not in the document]
rustonaut: Yep, I'm giving all the input to charity :stuck_out_tongue_winking_eye:

xayn-ai/src/ltr/list_net/ndlayers/dense.rs
```diff
@@ -110,15 +124,116 @@ where
...
+        // J = [
+        //  [ p_0 * s_0   p_1 * s_0 … ]
+        //  [ p_0 * s_1   p_1 * s_1 … ]
+        //  ⋮
+        // ]
+        // ```
+        //
+        // With dimensions [i, j].
+        let input = input.into_shape((weights_nr_rows, 1)).unwrap();
+        let partials = partials.into_shape((1, weights_nr_columns)).unwrap();
```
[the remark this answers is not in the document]
rustonaut: Yes they are. I documented the constraint.

Furthermore this is only used by `back_propagation` and only with arrays created there,
which are always created with the same kind of shapes so we are pretty much guaranteed
to catch any unexpected problems wrt. during tests.

xayn-ai/src/ltr/list_net/mod.rs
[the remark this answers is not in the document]
janpetschexain: makes sense regarding the `0`. we could still write it as a one-liner like `(count > 0).then(|| acc / count as f32)`.

rustonaut:
Just to prevent confusion the current branch head commit is just there temporary (since early today).

A problem with the bug is that it seems to be non-reproducible outside of github CI :shrug:

janpetschexain (approved):
sorry for the late approval

PR #144, opened 2021-07-09
Ty 1921 list net training random weights init

rustonaut:
Implemented initialization of ListNet based on random weights.

Weights are initialized using the He-Normal initialization and
biases are initialized  to 0.

This is based on PR #142, but can be separately reviewed (select to only show changes of the last commit).

### DIFF
--- a/xayn-ai/src/ltr/list_net/mod.rs
+++ b/xayn-ai/src/ltr/list_net/mod.rs
@@ -133,28 +133,31 @@ impl ListNet {
         let mut rng = rand::thread_rng();
 
         let dense1 = Dense::new(
-            he_normal_weights_init(&mut rng, Self::INPUT_NR_FEATURES, 48),
+            he_normal_weights_init(&mut rng, (Self::INPUT_NR_FEATURES, 48)),
             Array1::zeros((48,)),
             Relu,
         )
         .unwrap();
 
         let dense2 = Dense::new(
-            he_normal_weights_init(&mut rng, 48, 8),
+            he_normal_weights_init(&mut rng, (48, 8)),
             Array1::zeros((8,)),
             Relu,
         )
         .unwrap();
 
         let scores = Dense::new(
-            he_normal_weights_init(&mut rng, 8, 1),
+            he_normal_weights_init(&mut rng, (8, 1)),
             Array1::zeros((1,)),
             Linear,
         )
         .unwrap();
 
         let prob_dist = Dense::new(
-            he_normal_weights_init(&mut rng, Self::INPUT_NR_DOCUMENTS, Self::INPUT_NR_DOCUMENTS),
+            he_normal_weights_init(
+                &mut rng,
+                (Self::INPUT_NR_DOCUMENTS, Self::INPUT_NR_DOCUMENTS),
+            ),
             Array1::zeros((Self::INPUT_NR_DOCUMENTS,)),
             Softmax::default(),
         )
--- a/xayn-ai/src/ltr/list_net/mod.rs
+++ b/xayn-ai/src/ltr/list_net/mod.rs
@@ -126,8 +126,6 @@ impl ListNet {
     ///
     /// The weights are initialized using the He-Normal weight
     /// initializer, the biases are initialized to 0.
-    //TODO[maybe] run some tests to see weather or not Glorot or He
-    //            Uniform/Normal initialization works better for us.
     #[allow(dead_code)] //FIXME
     pub fn new_with_random_weights() -> Self {
         let mut rng = rand::thread_rng();
--- a/xayn-ai/src/ltr/list_net/mod.rs
+++ b/xayn-ai/src/ltr/list_net/mod.rs
@@ -9,6 +9,7 @@ use std::{
 use thiserror::Error;
 
 use ndarray::{s, Array1, Array2, ArrayView1, ArrayView2, Axis, Dimension, IntoDimension, Ix};
+use ndutils::he_normal_weights_init;
 
 use crate::Relevance;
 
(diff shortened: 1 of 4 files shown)


xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -121,6 +122,52 @@ impl ListNet {
         }
     }
 
+    /// Create a new ListNet instance with random weights.
+    ///
+    /// The weights are initialized using the He-Normal weight
+    /// initializer, the biases are initialized to 0.
+    //TODO[maybe] run some tests to see weather or not Glorot or He
```
janpetschexain, line 129:
```suggestion
    //TODO[maybe] run some tests to see whether or not Glorot or He
```

xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -121,6 +122,52 @@ impl ListNet {
...
+    //            Uniform/Normal initialization works better for us.
```
janpetschexain, line 130: afair, Glorot works better for layers with sigmoid activations and He for layers with relu activations. regarding the distribution, i once read a benchmark which showed slight benefits of uniform over normal, but that might be anecdotal and model dependent, in the end it shouldn't matter much.

xayn-ai/src/ltr/list_net/ndutils/mod.rs
```diff
@@ -52,6 +67,45 @@ pub fn kl_divergence(good_dist: ArrayView1<f32>, eval_dist: ArrayView1<f32>) ->
...
+/// This means for us `n_j` is the the number of rows of `W_j`.
+///
+/// Furthermore as we want to avoid exceedingly large values
+/// we truncate the normal distribution at 2σ.
+///
+/// Source:
+///
+/// - Website: https://www.cv-foundation.org/openaccess/content_iccv_2015/html/He_Delving_Deep_into_ICCV_2015_paper.html
+/// - Pdf: https://www.cv-foundation.org/openaccess/content_iccv_2015/papers/He_Delving_Deep_into_ICCV_2015_paper.pdf
+pub fn he_normal_weights_init<R>(rng: &mut R, nr_rows: Ix, nr_columns: Ix) -> Array2<f32>
```
janpetschexain, line 88: i'd say we shouldn't use `Ix` as the type (even though it works because it's just a `usize` alias), because it has a very specific meaning as the array dimensionality in `ndarray`. the arrays in this case always have dimension 2 and we actually want to express the shape/size of each dimension.
```suggestion
pub fn he_normal_weights_init<R>(rng: &mut R, nr_rows: usize, nr_columns: usize) -> Array2<f32>
```

xayn-ai/src/ltr/list_net/ndutils/mod.rs
```diff
@@ -258,4 +312,42 @@ mod tests {
...
+        let nr_weights = weights.len() as f32;
+        let prob_1std = c_1std as f32 / nr_weights;
+        let prob_2std = c_2std as f32 / nr_weights;
+
+        // Probabilities of a weight being in +-1std or +-2std corrected
+        // wrt. the truncating of values outside of +-2std. We accept this
+        // to be true if the found percentage is around +-5% of the expected
+        // percentage.
+        assert_approx_eq!(f32, prob_1std, 0.713_389_1, epsilon = 0.05);
+        assert_approx_eq!(f32, prob_2std, 0.284_518_84, epsilon = 0.05);
```
janpetschexain, line 351: it shouldn't make much of a difference, but i get slightly different probablilites ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=fa30ae9782fec47cf21e63c3d746f8e6)):
```suggestion
        assert_approx_eq!(f32, prob_1std, 0.715_232_8, epsilon = 0.05);
        assert_approx_eq!(f32, prob_2std, 0.284_767_2, epsilon = 0.05);
```

xayn-ai/src/ltr/list_net/mod.rs
rustonaut: I did a bit of ad-hoc research when deciding what to use and also came to the same result. But also found enough sources which:

- Show that the idea when Glorot or He is better don't always work out.
- The difference of Uniform and Normal is very little and supposedly papers which show one is better then the other are often not reproducible if just slightly different networks are used.

Either way TensorFlows default is Glorot Uniform and the soundgarden implementation does not change this, but when I train on the 500_000 user samples I always get a slightly better evaluation cost. E.g. 0.312... instead of 0.319... .

Now that I can use the GPU systems I could try running all for combinations (not in this PR, but based on my current working branch) it would be pretty simple.

xayn-ai/src/ltr/list_net/ndutils/mod.rs
rustonaut: Ix stands for "array index type" (says the `ndarray` documentation) it's the type used for indexing and describing the length of dimensions. `Ix2` e.g. stands for "two array index type values" (wrapped into a `Dim`).

Similar the passed in `nr_rows`, `nr_columns` are used as `(nr_rows, nr_columns)` to create a new array and the method in `ndarray` using that tuple is based onto `impl IntoDimension for (Ix, Ix)` similar if you index into a array it's based on e.g. `&[Ix]`, `(Ix,Ix)`, etc.

But then the returned shape is `&[usize]` and slicing also doesn't use `Ix` :thinking: .

So while I see where you are coming from (Ix being 1,2,3 dimensions) I would also argue that:

- The usage here is in the boundaries of what it's supposed to be used for, especially given that it's mainly used
  to create a `(Ix,Ix)` tuple to create the shape.
- `ndarray` isn't always fully consistent about the usage of `Ix` and `usize`.

Either way I changed it to use `dim: impl IntoDimension<Dim=Ix2>` instead, so it now mirrors all the
`Array` constructors  (like `from_...` or `zeros`) but limited to `Ix2`.

xayn-ai/src/ltr/list_net/ndutils/mod.rs
rustonaut: I did use much less precise numbers for the calculation and calculated it using ipython and then converted it to f32 (as the difference in precision shouldn't matter given the sample size and a epsilon of 0.05).

I fixed it to use your numbers.

janpetschexain (approved):

PR #146, opened 2021-07-15
TY-1944 add qambert example for rubert

janpetschexain:
**References**

- [TY-1944]

**Summary**

- adjust the smbert example to provide for multiple kinds of mberts

[TY-1944]: https://xainag.atlassian.net/browse/TY-1944


acrrd:
What about having only one example and decide what to load depending on `std::env::args()[1]`?

acrrd (approved):

PR #147, opened 2021-07-20
TY-1953 compare onnx vs tract

janpetschexain:
**References**

- [TY-1953]
- [old python script](https://github.com/xaynetwork/xain_ai/tree/dev/xaynai_flutter/python)

**Summary**

- expose `rubert::model::kinds`
- add `validate` binary: configuration, models, data processing & error computations

**Todo**

- the ted talks file is already stored somewhere in our cloud storage because of the old validation script (i haven't looked for it yet, at least it used to be there), should we add this to our download script?

[TY-1953]: https://xainag.atlassian.net/browse/TY-1953


rubert/src/bin/validate.rs
```diff
@@ -0,0 +1,356 @@
...
+                    session: None,
+                    _pinned: PhantomPinned,
+                };
+                if let Self::OnnxMBert {
+                    ref environment,
+                    ref mut session,
+                    ..
+                } = pipeline
+                {
+                    *session = unsafe {
```
acrrd, line 146: Can't we build `session` after `environment` and then create the `OnnxMBert` pipeline?

rubert/src/bin/validate.rs
```diff
@@ -0,0 +1,356 @@
...
+        let talks = config.data.talks;
+        let skip = match config.data.range.start_bound() {
+            Bound::Included(start) => *start,
+            Bound::Excluded(start) => start + 1,
+            Bound::Unbounded => 0,
+        };
+        let take = match config.data.range.end_bound() {
+            Bound::Included(end) => end + 1,
+            Bound::Excluded(end) => *end,
+            Bound::Unbounded => 2467,
```
acrrd, line 235: Shouldn't this be the number of lines in `talks`?

rubert/src/bin/validate.rs
```diff
@@ -0,0 +1,356 @@
...
+        } - skip;
+        let source = Pipeline::build(&config.tokenizer, &config.source);
+        let target = Pipeline::build(&config.tokenizer, &config.target);
+        let errors = Array1::zeros(11);
```
acrrd, line 239: Why 11?

rubert/src/bin/validate.rs
```diff
@@ -0,0 +1,356 @@
...
+        (norms.is_finite() && norms > 0.0)
+            .then(|| (source.deref() * target.deref() / norms).sum())
+            .unwrap_or_default()
+    }
+
+    /// Computes various errors between source and target embeddings based on the chosen ted talks.
+    fn validate(&mut self) -> &mut Self {
+        let mut reader = Reader::from_path(self.talks).unwrap();
+        let progress = ProgressBar::new(self.take as u64);
+        let mut errors = Array2::<f32>::zeros((330644, 5));
```
acrrd, line 298: why `330644`?

rubert/src/bin/validate.rs
janpetschexain: this very much depends on how we structure the whole binary. unfortunately the `Session` contains a `&Environment`which leaves us some options:
- create a self-referential struct which requires unsafe code
- have an environment and session separately in the main function which ruins the abstraction layers
- create a session everytime the model is run which is terribly slow

i get why the `onnxruntime` binds the session to the livetime of the environment, but it is a really cumbersome design. also, i noticed that i forgot to actually pin the environment even though i not unpin, but it accidentally worked probably because the memory wasn't moved. anyways i fixed that, updated the safety comment and had to add some more hackery: we can't pin the session together with the environment, because running it requires a mutable reference, which is again due to the `onnxruntime` :/

rubert/src/bin/validate.rs
janpetschexain: interestingly, the number of lines in the csv file don't correspond to the number of records, probably due to some additional line breaks because of extra long talks.

rubert/src/bin/validate.rs
janpetschexain: one for the actual number of sentences, and two each for the mean and standard deviation for each of the five errors.

rubert/src/bin/validate.rs
janpetschexain: this is the total number of sentences in all talks.

i added comments in the code to explain all the magic numbers above.

acrrd (approved):
Maybe in the future we could make this to be able to get an arbitrary input but for the moment is fine.

PR #148, opened 2021-07-20
Ty 1931 list net training based on dev tool

rustonaut:
Adds support for training on soundgarden user data-frames.

Based on #144 .

Adds following sub-commands to `dev-tool list-net`:

- `convert` prepare samples for training (by converting them from soundgarden user data-frames to a sampels files)
- `train` trains ListNet
- `inspect` inspect ListNet parameter files and/or sample files.


rustonaut (commented):
Self review to check code from a different perspective.

rustonaut:
(I realized I need to update the data on S3, so because of this and the NaN problem some tests will fail.)

xayn-ai/src/ltr/mod.rs
```diff
@@ -116,6 +119,86 @@ impl LtrSystem for ConstLtr {
     }
 }
 
+pub type OwnedSample = (Array2<f32>, Array1<f32>);
+
+/// Creates training data for ListNet from a users history.
+pub fn list_net_training_data_from_history(
+    mut history: &[DocumentHistory],
+) -> Result<Vec<OwnedSample>, Error> {
```
rustonaut, line 127: This implementation is far from perfect, but will be replaced anyway in a follow-up PR.

rustonaut (commented):
Self review + comments for other reviewers.

janpetschexain:
is there a way to split this pr into smaller chunks? 3.5k changed loc is quite a lot oO

PR #152, opened 2021-07-27
TY-1952 sync coi ids

finiteprods:
**Summary**

a follow up to #143 for synchronization to support global consistency of CoI ids. 

- refactors `CoiId(usize)` to `CoiId(Uuid)`, and various functions (e.g. methods of `CoiPoint`) to hide the underlying type.
- a new id is now assigned to a CoI whenever it is updated locally.
- duplicate CoIs (indicated by clashing ids) are removed during synchronization.

### DIFF
--- a/xayn-ai/src/coi/system.rs
+++ b/xayn-ai/src/coi/system.rs
@@ -232,7 +232,7 @@ mod tests {
         },
         reranker::systems::CoiSystem as CoiSystemTrait,
         to_vec_of_ref_of,
-        utils::mock_coiid,
+        utils::mock_coi_id,
     };
 
     pub(crate) fn create_data_with_mab(
@@ -251,7 +251,7 @@ mod tests {
                 },
                 qambert: QAMBertComponent { similarity: 0.5 },
                 coi: CoiComponent {
-                    id: mock_coiid(1),
+                    id: mock_coi_id(1),
                     pos_distance: 0.1,
                     neg_distance: 0.1,
                 },
@@ -358,7 +358,7 @@ mod tests {
 
     #[test]
     fn test_shift_coi_point() {
-        let coi = PositiveCoi::new(mock_coiid(0), arr1(&[1., 1., 1.]).into());
+        let coi = PositiveCoi::new(mock_coi_id(0), arr1(&[1., 1., 1.]).into());
         let embedding = arr1(&[2., 3., 4.]).into();
 
         let updated_coi = CoiSystem::default().shift_coi_point(&embedding, &coi.point);
@@ -409,7 +409,7 @@ mod tests {
             .compute_coi_for_embedding(&embedding, &user_interests)
             .unwrap();
 
-        assert_eq!(coi_comp.id, mock_coiid(2));
+        assert_eq!(coi_comp.id, mock_coi_id(2));
         assert_approx_eq!(f32, coi_comp.pos_distance, 4.8904557);
         assert_approx_eq!(f32, coi_comp.neg_distance, 8.1273575);
     }
@@ -428,7 +428,7 @@ mod tests {
             .compute_coi_for_embedding(&embedding, &user_interests)
             .unwrap();
 
-        assert_eq!(coi_comp.id, mock_coiid(2));
+        assert_eq!(coi_comp.id, mock_coi_id(2));
         assert_approx_eq!(f32, coi_comp.pos_distance, 4.8904557);
         assert_approx_eq!(f32, coi_comp.neg_distance, f32::MAX, ulps = 0);
     }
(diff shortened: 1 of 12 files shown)


xayn-ai/src/data/mod.rs
```diff
@@ -1,14 +1,18 @@
...
 use crate::embedding::utils::Embedding;
 
 // Hint: We use this id new-type in FFI so repr(transparent) needs to be kept
 #[repr(transparent)]
-#[derive(Debug, PartialEq, Eq, Hash, Clone, Copy, Serialize, Deserialize, PartialOrd, Ord)]
-pub struct CoiId(pub usize);
+#[derive(
+    Debug, PartialEq, Eq, Hash, Clone, Copy, PartialOrd, Ord, Serialize, Deserialize, From,
+)]
+pub struct CoiId(pub Uuid);
```
janpetschexain, line 15: should we make the inner field private since we have the newly derived `From` impl now?
```suggestion
pub struct CoiId(Uuid);
```

xayn-ai/src/coi/system.rs
```diff
@@ -356,7 +358,7 @@ mod tests {
 
     #[test]
     fn test_shift_coi_point() {
-        let coi = PositiveCoi::new(0, arr1(&[1., 1., 1.]).into());
+        let coi = PositiveCoi::new(Uuid::nil().into(), arr1(&[1., 1., 1.]).into());
```
janpetschexain, line 361: i'd suggest to use the mocked uuid here as well
```suggestion
        let coi = PositiveCoi::new(mock_uuid(0).into(), arr1(&[1., 1., 1.]).into());
```

xayn-ai/src/context.rs
```diff
@@ -109,7 +109,7 @@ mod tests {
                 smbert: SMBertComponent { embedding },
                 qambert: QAMBertComponent { similarity },
                 coi: CoiComponent {
-                    id: CoiId(0),
+                    id: Uuid::nil().into(),
```
janpetschexain, line 112:
```suggestion
                    id: mock_uuid(0).into(),
```

janpetschexain (approved):
lgtm

since we are using `mock_uuid(id).into()` in so many places now, we might as well add a wrapper for it.
```rust
#[cfg(test)]
pub(crate) fn mock_coiid(sub_id: usize) -> CoiId {
    mock_uuid(sub_id).into()
}
```

xayn-ai/src/data/mod.rs
finiteprods: yes makes sense to hide it

finiteprods: yes makes sense to hide it

xayn-ai/src/utils.rs
```diff
@@ -78,6 +78,12 @@ pub(crate) fn mock_uuid(sub_id: usize) -> uuid::Uuid {
     uuid::Uuid::from_u128(BASE_UUID | (sub_id as u128))
 }
 
+/// Creates a CoI id from a mock UUID.
+#[cfg(test)]
+pub(crate) fn mock_coiid(sub_id: usize) -> crate::CoiId {
```
rustonaut, line 83: Minor nitpicking: I think it should be `mock_coi_id` given that it's also `CoidId` and not `Coiid`.

xayn-ai/src/reranker/sync.rs
```diff
@@ -51,41 +50,21 @@ impl SyncData {
...
-    let max_local_id = locals.iter().map(|coi| coi.id().0).max().unwrap_or(0);
-    remotes
-        .iter_mut()
-        .for_each(|coi| coi.set_id(max_local_id + coi.id().0));
-
+/// Appends `remotes` to `locals`.
+///
+/// Remote CoIs with ids clashing with any of the local CoIs are removed.
+fn append_cois<C: CoiPoint>(locals: &mut Vec<C>, remotes: &mut Vec<C>) {
+    remotes.retain(|rem| !locals.iter().any(|loc| loc.id() == rem.id()));
```
rustonaut, line 68: Do I assume correctly that we remove colliding Cois because we assume matches means that they had been synced before? (Assuming that UUID collisions in general don't happen.)

Besides that how many ids will this have in general. If they are many, we might want to use a "append to hash-set" schema instead of an retain + iter schema which is O(n^2).

rustonaut (commented):
On nitpicking and one small question.

xayn-ai/src/reranker/sync.rs
finiteprods: yes, it means they have been synced before and neither has been updated since the last sync, assuming uuid collisions are astronomically unlikely.

how often this happens would depend on usage I suppose, e.g. if there's little activity between syncs, not many CoIs will be updated, the more likely there will be dupes like this. using something like a hash set would be more optimal in general yes - this current version was just to get something simple working.

PR #154, opened 2021-07-29
TY-1897 expose sync methods to ffi

finiteprods:
**Summary**

previous related: #143, #152.
- exposes the functions `syncdata_bytes`, `synchronize` to ffi layers and dart binding.
- corresponding tests.
- handles panic when empty passed to `synchronize` (see `sync.rs`).
- updates changelog.

### DIFF
--- a/bindings/dart/lib/src/common/reranker/ai.dart
+++ b/bindings/dart/lib/src/common/reranker/ai.dart
@@ -69,7 +69,7 @@ class XaynAi {
   Uint8List syncdataBytes() => throw UnsupportedError('Unsupported platform.');
 
   /// Synchronizes the internal data of the reranker with another.
-  void synchronize([Uint8List? serialized]) =>
+  void synchronize(Uint8List serialized) =>
       throw UnsupportedError('Unsupported platform');
 
   /// Frees the memory.
--- a/bindings/dart/lib/src/common/reranker/ai.dart
+++ b/bindings/dart/lib/src/common/reranker/ai.dart
@@ -65,6 +65,13 @@ class XaynAi {
   /// Retrieves the analytics which were collected in the penultimate reranking.
   Analytics? analytics() => throw UnsupportedError('Unsupported platform.');
 
+  /// Serializes the synchronizable data of the reranker.
+  Uint8List syncdataBytes() => throw UnsupportedError('Unsupported platform.');
+
+  /// Synchronizes the internal data of the reranker with another.
+  void synchronize([Uint8List? serialized]) =>
+      throw UnsupportedError('Unsupported platform');
+
   /// Frees the memory.
   void free() => throw UnsupportedError('Unsupported platform.');
 }
--- a/xayn-ai-ffi-c/src/reranker/ai.rs
+++ b/xayn-ai-ffi-c/src/reranker/ai.rs
@@ -632,7 +632,7 @@ mod tests {
     }
 
     #[test]
-    fn test_syncdata_bytes_synchronize() {
+    fn test_syncdata_bytes() {
         let smbert_vocab = TestFile::smbert_vocab();
         let smbert_model = TestFile::smbert_model();
         let qambert_vocab = TestFile::qambert_vocab();
@@ -641,7 +641,7 @@ mod tests {
         let db = TestDb::default();
         let mut error = CError::default();
 
-        let mut xaynai = unsafe {
+        let xaynai = unsafe {
             xaynai_new(
                 smbert_vocab.as_ptr(),
                 smbert_model.as_ptr(),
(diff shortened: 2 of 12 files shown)


xayn-ai-ffi-c/src/reranker/ai.rs
```diff
@@ -316,6 +339,58 @@ pub unsafe extern "C" fn xaynai_analytics(
...
+/// - An unexpected panic happened.
+///
+/// In case of a [`CCode::Panic`] the Xayn AI must not be accessed anymore and should be dropped via
+/// [`xaynai_drop()`]. The last known valid state can be restored by the caller via [`xaynai_new()`]
+/// with a previously serialized reranker database obtained from [`xaynai_serialize()`].
+///
+/// # Safety
+/// The behavior is undefined if:
+/// - A non-null `xaynai` doesn't point to memory allocated by [`xaynai_new()`].
+/// - A non-null `error` doesn't point to an aligned, contiguous area of memory with a [`CError`].
```
janpetschexain, line 382: are there any additional safety concerns for the `CBytes` next to the usual ones of the `CBoxedSlice`? (probably not?)
```suggestion
/// - The safety constraints of [`CBoxedSlice`] are violated.
/// - A non-null `error` doesn't point to an aligned, contiguous area of memory with a [`CError`].
```

xayn-ai-ffi-c/src/reranker/ai.rs
```diff
@@ -1158,6 +1267,21 @@ mod tests {
...
+    #[test]
+    fn test_ai_null_sync_bytes() {
+        let mut error = CError::default();
+
+        let invalid = None;
+        assert!(unsafe { xaynai_syncdata_bytes(invalid, error.as_mut_ptr()) }.is_none());
+        assert_eq!(error.code, CCode::AiPointer);
+        assert_eq!(
+            error.message.as_ref().unwrap().as_str(),
+            "Failed to serialize sync data: The ai pointer is null"
```
janpetschexain, line 1279:
```suggestion
            "Failed to serialize sync data: The ai pointer is null",
```

xayn-ai-ffi-c/src/reranker/ai.rs
```diff
@@ -316,6 +339,58 @@ pub unsafe extern "C" fn xaynai_analytics(
...
+/// with a previously serialized reranker database obtained from [`xaynai_serialize()`].
+///
+/// # Safety
+/// The behavior is undefined if:
+/// - A non-null `xaynai` doesn't point to memory allocated by [`xaynai_new()`].
+/// - A non-null `error` doesn't point to an aligned, contiguous area of memory with a [`CError`].
+#[no_mangle]
+pub unsafe extern "C" fn xaynai_synchronize(
+    xaynai: Option<&mut CXaynAi>,
+    bytes: CBytes,
```
janpetschexain, line 386: usually fat pointers won't work for the dart wrapper. since we get a thin pointer from `xaynai_syncdata_bytes()` anyways, i'd suggest to use it as input for this method too.
```suggestion
    bytes: Option<Box<CBytes>>,
```

alternatively this function could accept non-owned bytes, which probably makes more sense as this is likely to be called on a different device than the bytes getter.
```suggestion
    bytes: Option<&CBytes>,
```

finiteprods: thanks! I've gone with the latter suggestion.

bindings/dart/lib/src/common/reranker/ai.dart
```diff
@@ -65,6 +65,13 @@ class XaynAi {
   /// Retrieves the analytics which were collected in the penultimate reranking.
   Analytics? analytics() => throw UnsupportedError('Unsupported platform.');
 
+  /// Serializes the synchronizable data of the reranker.
+  Uint8List syncdataBytes() => throw UnsupportedError('Unsupported platform.');
+
+  /// Synchronizes the internal data of the reranker with another.
+  void synchronize([Uint8List? serialized]) =>
```
janpetschexain, line 72: since this argument is marked as optional, does it make sense to call the method without any arguments? i would think that we could make this required like in the reranker of `xayn_ai`:
```suggestion
  void synchronize(Uint8List serialized) =>
```
(the c ffi still needs to take an `Option<&CBytes>`, as this represents any thin pointer to a `CBytes` fat pointer, where the former theoretically could also be null.)

xayn-ai-ffi-wasm/src/ai.rs
```diff
@@ -648,7 +751,7 @@ mod tests {
             QAMBERT_VOCAB,
             QAMBERT_MODEL,
             LTR_MODEL,
-            Some(Box::new([2, 3, 4])),
+            Some(Box::new([u8::MAX])),
```
rustonaut, line 754: I have no complain against the change.

But out of interest why was this changed?

rustonaut (approved):
lgtm

xayn-ai-ffi-wasm/src/ai.rs
finiteprods: it better mirrors how this test is implemented in other places, and it's the "minimal" example - we just need the first byte to be an invalid version, so `u8::MAX` makes it comfortably so. then `2,3` or anything else is unnecessary.

xayn-ai-ffi-c/src/reranker/ai.rs
```diff
@@ -400,6 +481,21 @@ mod tests {
...
+    struct TestSyncdata(CBytes);
+
+    impl Default for TestSyncdata {
+        fn default() -> Self {
+            Self(vec![0; 17].into_boxed_slice().into())
+        }
+    }
+
+    impl TestSyncdata {
+        #[allow(clippy::clippy::unnecessary_wraps)]
```
janpetschexain, line 493: seems like the autocompletion added a double clippy
```suggestion
        #[allow(clippy::unnecessary_wraps)]
```

janpetschexain (approved):

xayn-ai-ffi-c/src/reranker/ai.rs
finiteprods: thanks, changed it. funny it still worked fine..

PR #155, opened 2021-08-02
TY-1978 update metadata

Robert-Steiner:
**Ticket**

- [TY-1978]

**Summary**

- renamed `suffix` to `urlSuffix`
- moved the checksum method into a new `Checksum` class
- added a `fragments` field to the `Asset` class
  - each `Fragment` consists of a `checksum` and a `urlSuffix`
- updated the example

[TY-1978]: https://xainag.atlassian.net/browse/TY-1978

PR #156, opened 2021-08-03
TY-1980 split and generate assets metadata

Robert-Steiner:
**Ticket:** 

[TY-1980]

**Summary**

- added a `generate_assets_metadata.sh` script that splits the models into fragments and generates the assets manifest
	- the script replaces the `assets.json.tmpl` file and works locally and on the CI 
- simplified `Makefile.toml` generate assets is now only called once when building the web version
- updated and cleaned up the build asset job

**Please check if the script runs on your local machine!**

Besides `gomplate` you will also need [`jq`](https://github.com/stedolan/jq).

## Example code for loading the assets on the web version 

**bindings/dart/example/lib/data_provider/web.dart**

```dart
import 'dart:typed_data' show ByteBuffer, BytesBuilder, Uint8List;

/// Prepares and returns the data that is needed to init [`XaynAi`].
Future<SetupData> getInputData() async {
  final fetched = <AssetType, Uint8List>{};

  for (var asset in getAssets().entries) {
    if (asset.value.fragments.isEmpty) {
      final path = joinPaths([_baseAssetUrl, asset.value.urlSuffix]);
      final data = await _fetchAsset(path, asset.value.checksum.checksumSri);
      fetched.putIfAbsent(asset.key, () => data);
    } else {
      final builder = BytesBuilder(copy: false);
      for (var fragment in asset.value.fragments) {
        final path = joinPaths([_baseAssetUrl, fragment.urlSuffix]);
        final part = await _fetchAsset(path, fragment.checksum.checksumSri);
        builder.add(part);
      }
      fetched.putIfAbsent(asset.key, () => builder.takeBytes());
    }
  }

  return SetupData(fetched);
}
```

Based on #155 

[TY-1980]: https://xainag.atlassian.net/browse/TY-1980


assets_manifest.json
```diff
@@ -0,0 +1,41 @@
...
+    {
+      "dart_enum_name": "smbertVocab",
+      "url_suffix": "smbert_v0000/vocab.txt",
+      "path": "data/smbert_v0000/vocab.txt"
+    },
+    {
+      "dart_enum_name": "smbertModel",
+      "url_suffix": "smbert_v0000/smbert.onnx",
+      "path": "data/smbert_v0000/smbert.onnx",
+      "chunk_size": "10MB"
```
acrrd, line 12: I think we could use 11 here, so we can have fragments that are more similar in size (the last chunk will be ~8mb instead of ~2)

generate_assets_metadata.sh
```diff
@@ -0,0 +1,154 @@
...
+        else
+            echo "Requires the GNU version of 'split'. Use 'brew install coreutils' to install it."
+            exit 1
+        fi
+    else
+        SPLIT="split"
+    fi
+else
+    if [ $RUNNER_OS == "macOS" ]; then
+        brew install coreutils
```
acrrd, line 29: Do we want to do it here instead of the CI and maybe caching it? This will also allow us to delete this branch an leave only the previous check.

generate_assets_metadata.sh
```diff
@@ -0,0 +1,154 @@
...
+    flutter format bindings/dart/lib/src/common/reranker/assets.dart
+}
+
+calc_checksum() {
+    echo $(shasum -a 256 $1 | awk '{ print $1 }')
+}
+
+gen_ai_assets_metadata() {
+    local ASSET_MANIFEST=$1
+    local TMP_FILE=$OUT_DIR/tmp_file
```
acrrd, line 74: We can also use `mktemp` to create a temporary file.

generate_assets_metadata.sh
acrrd: Do we need to run this on the mac ci? This is only needed when we push to the release repo and can be done on the ubuntu instance.

generate_assets_metadata.sh
```diff
@@ -0,0 +1,154 @@
...
+    local CHUNKS_DIR=$OUT_DIR/chunks
+
+    $(rm -rf $CHUNKS_DIR || true) && mkdir $CHUNKS_DIR
```
acrrd, line 77: Should we also do the same with `TMP_FILE`? (If we don't use `mktemp`).

generate_assets_metadata.sh
```diff
@@ -0,0 +1,154 @@
...
+
+            local FRAGMENTS=""
+            for CHUNK_PATH in $(find ${CHUNKS_DIR}/${ASSET_VERSION} -name "${ASSET_FILENAME}_*" | sort -n); do
+                local FRAGMENT_CHECKSUM=$(calc_checksum $CHUNK_PATH)
+                local FRAGMENT_FILENAME=$(basename $CHUNK_PATH)
+                local FRAGMENT="{\"path\": \"$CHUNK_PATH\", \"url_suffix\": \"${ASSET_VERSION}/${FRAGMENT_FILENAME}\", \"checksum\": \"$FRAGMENT_CHECKSUM\"}"
+                if [ -z "${FRAGMENTS}" ]; then
+                    FRAGMENTS="$FRAGMENT"
+                else
+                    FRAGMENTS="$FRAGMENTS, $FRAGMENT"
```
acrrd, line 101:
```suggestion
                    FRAGMENTS+=", $FRAGMENT"
```

generate_assets_metadata.sh
Robert-Steiner: Unfortunately yes, we need it when we build the iOS flutter example

generate_assets_metadata.sh
Robert-Steiner: good idea 👍

generate_assets_metadata.sh
Robert-Steiner: I moved the installation of `coreutils` into the action

generate_assets_metadata.sh
```diff
@@ -0,0 +1,155 @@
+#!/bin/bash
```
janpetschexain, line 1: since we have more and more scripts in the repo root, should we move them into a dedicated directory in another pr?

generate_assets_metadata.sh
```diff
@@ -0,0 +1,155 @@
...
+# ./generate_assets_metadata <path of assets_manifest.json> [<wasm version>] [<path of the wasm output directory>]
+
+set -e
+
+OUT_DIR="out"
+ASSETS_METADATA_PATH=$OUT_DIR/assets_metadata.json
+
+if [ -z ${GITHUB_ACTIONS} ]; then
+    if [[ "$OSTYPE" == "darwin"* ]]; then
+        if command gsplit --version /dev/null; then
```
janpetschexain, line 20: this could also be written like the following to avoid the need to redirect the output
```suggestion
        if [ -x  "$(command -v gsplit)" ]; then
```
similarly below.

generate_assets_metadata.sh
Robert-Steiner: sounds good 👍
I will create a task for that

generate_assets_metadata.sh
```diff
@@ -0,0 +1,156 @@
...
+        if [ -x "$(command -v gsplit)" ]; then
+            SPLIT="gsplit"
+        else
+            echo "Cannot find 'gsplit'."
+            exit 1
+        fi
+    else
+        SPLIT="split"
+    fi
+fi
```
acrrd, line 40: Can we remove the outer if? 
```suggestion
if [[ "$OSTYPE" == "darwin"*  || $RUNNER_OS == "macOS" ]]; then
        if [ -x "$(command -v gsplit)" ]; then
            SPLIT="gsplit"
        else
            echo "Requires the GNU version of 'split'. Use 'brew install coreutils' to install it."
            exit 1
        fi
else
        SPLIT="split"
fi
```

acrrd (approved):

PR #157, opened 2021-08-04
TY-1991 parallelization benchmarks

janpetschexain:
**References**

- [TY-1991]
- [rust book on parallelism](https://doc.rust-lang.org/book/ch16-00-concurrency.html)
- [rustonomicon on parallelism](https://doc.rust-lang.org/nomicon/concurrency.html)

**Use case constraints**

- MBert pipeline and sequence only taken by shared reference to produce an embedding
- MBert pipelines can run independently and take the same batch of sequences as input
- each MBert pipeline can run the sequences/samples of the batch independently

**Memory sharing vs message passing**

- parallelization of batch looping is straight forward (e.g. with `rayon`), parallelization of pipelines is also kind of easy due to shared references; passing messages would probably require a clone of the pipeline per channel (maybe `Arc`ing works as well?)
- little change needed, ios & android & wasm compile successfully with `rayon` (i haven't run them though, wasm may need additional setup via [wasm-bindgen-rayon](https://crates.io/crates/wasm-bindgen-rayon)), but some targets may need conditional compilation/fallbacks ([like safari for wasm](https://webassembly.org/roadmap/)); channels will probably work without issues on any platform but multithreading still has the same constraints as in the shared memory case

**Further parallelization opportunities**

- remove need for parallelization of the tokenization by running the tokenization only once (requires same tokenizer configuration and vocabulary which is currently the case but not enforced in the code and the data archives), but this may be done later as the tokenization only accounts for roughly 5-7% of the pipeline (less than 1% for the pooler, rest for the model)

**Potential speed improvements**

- checked on my mac with four cores, depending on processor load we can achieve almost a speedup up to the number of available cores in the fully parallized setting
- SMBert sequential -> parallel: 3.5-3.8x
- QAMBert sequential -> parallel: 3.3-3.7x
- SMBert & QAMBert sequential -> join: 1.7-1.9x
- SMBert & QAMBert sequential -> parallel: 3.2-3.5x
- SMBert & QAMBert sequential -> join parallel: 3.8-3.9x

[TY-1991]: https://xainag.atlassian.net/browse/TY-1991

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1477,6 +1477,7 @@ dependencies = [
  "indicatif",
  "ndarray",
  "onnxruntime",
+ "rayon",
  "rubert-tokenizer",
  "serde",
  "thiserror",
--- a/rubert/Cargo.toml
+++ b/rubert/Cargo.toml
@@ -22,9 +22,10 @@ criterion = { version = "0.3.5", features = ["html_reports"], optional = true }
 csv = { version = "1.1.6", optional = true }
 indicatif = { version = "0.16.2", optional = true }
 onnxruntime = { version = "0.0.13", optional = true }
+rayon = { version = "1.5.1", optional = true }
 
 [features]
-bench = ["criterion", "onnxruntime"]
+bench = ["criterion", "onnxruntime", "rayon"]
 validate = ["csv", "indicatif", "onnxruntime"]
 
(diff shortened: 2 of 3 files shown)


Robert-Steiner (approved):
🚀

PR #159, opened 2021-08-04
ListNet training with soundgarden; Split out all changes  to the xayn-ai crate from original PR (part 1).

rustonaut:
WARNING:

This is splitting out all changes to `xayn-ai` into it's own PR, this means this will add some trait implementations and make some thinks public where it seems as it might not be necessary. 

Another side note: The `fn list_net_training_data_from_history` function will be completely overwritten in a follow up PR (not part of this splitting) so while it's not grate I also don't want to spend time on it until it's re-written to support our needs which will change it and it's signature fundamentally.


janpetschexain (approved):
lgtm

xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -1037,101 +1135,477 @@ mod tests {
...
+
+        let relevances = vec![Low, Low, Medium, Medium, Low, Medium, High, Low, High, High];
+        let data_frame = (inputs, prepare_target_prob_dist(&relevances).unwrap());
+
+        let training_data = vec![data_frame.clone(), data_frame.clone(), data_frame.clone()];
+        let test_data = vec![data_frame];
+        let nr_epochs = 5;
+        let batch_size = 1;
+
+        // The NaN bug somehow only appear on the first execution so uncommenting this
```
acrrd, line 1280: Is this still relevant?

xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -1341,4 +1793,24 @@ mod tests {
...
+        assert_approx_eq!(f32, list_net.dense2.bias(), list_net2.dense2.bias());
+        assert_approx_eq!(f32, list_net.scores.weights(), list_net2.scores.weights());
+        assert_approx_eq!(f32, list_net.scores.bias(), list_net2.scores.bias());
+        assert_approx_eq!(
+            f32,
+            list_net.prob_dist.weights(),
+            list_net2.prob_dist.weights()
+        );
+        assert_approx_eq!(f32, list_net.prob_dist.bias(), list_net2.prob_dist.bias());
+    }
```
acrrd, line 1815: This file si pretty huge, can we move the tests to a directory `tests` here and have something like `tests/ltr`, `tests/training`, or some other split.

acrrd (approved):
Should we think about moving the domain reranker to it's own create like rubert?

xayn-ai/src/ltr/list_net/mod.rs
acrrd: In this or another pr.

rustonaut: Sure, I should be able to put it in this PR.

rustonaut:
> Should we think about moving the domain reranker to it's own create like rubert?

I considered that, too. But I would want to wait until the current string of PRs is done. 
I fear doing it now might cause a bunch of headaches wrt. the follow up PRs.

acrrd:
> > Should we think about moving the domain reranker to it's own create like rubert?
> 
> I considered that, too. But I would want to wait until the current string of PRs is done.
> I fear doing it now might cause a bunch of headaches wrt. the follow up PRs.

Absolutely, was more a general question, not something to do now.

xayn-ai/src/ltr/mod.rs
```diff
@@ -116,6 +119,86 @@ impl LtrSystem for ConstLtr {
...
+
+        let last_query_results_id = {
+            let last = history.last().unwrap();
+            (last.session, last.query_count)
+        };
+        let start_of_last_query = history
+            .iter()
+            .rposition(|doc| (doc.session, doc.query_count) != last_query_results_id)
+            .map(|last_before_last_query_idx| last_before_last_query_idx + 1)
+            .unwrap_or_default();
```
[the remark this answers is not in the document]
rustonaut: It's the id of the last results(1) of a query, hence `last_query_results_id`.

(1): Note the `s` the `last_query_result_id` would be `(session_id, query_count, position)`.

I agree that the terminology is confusing, I mean query is sometimes used for the specific query done by the user and sometimes refers to the results of that query.  So I will do the proposed changes.

xayn-ai/src/ltr/list_net/mod.rs
```diff
@@ -1017,6 +1113,7 @@ mod tests {
     }
 
     struct VecDataSource {
+        batch_size: usize,
```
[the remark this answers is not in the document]
rustonaut: I would change it, if I hadn't already done so (kinda) in a follow up PR :wink: .

Mainly I change the `DataSource` trait to not set the batch size when calling `reset` but when creating a instance implementing the trait. Which is generally a much more clean design.

xayn-ai/src/ltr/list_net/mod.rs
rustonaut: I hope not, we didn't really found the bug but happened to work around it.

And while we kinda expect it to be caused by QEMU, it's possible that it is a miscompilation of rust
where rust relies on some kind of implementation detail which happens to work this way on most
hardware but happened to not be implemented that way in QEMU. 

Through I guess I can remove it anyway.

PR #160, opened 2021-08-04
TY-1931 ListNet training with soundgarden: Split PR, part 2. Add samples preparation CLI.

rustonaut:
This adds the `dev-tool list-net convert` sub-command which can be used to convert a folder with soundgarden user data frame csv files to a `<name>.samples` file (which is a bincode serialized version of `InMemorySamples`).

For the 500_000 samples with are ~5.4GiB unpacked on disk this will produce a 875MiB file. Which if compiled in release mode will load in ~1s (or ~50s if not in release mode).

The `<name>.samples` files are used for training using soundgarden data, both with the classical training which is added in part 3 and the XaynNet emulation training which will be added in a followup PR.

PS:
Also don't forget that each split out PR will contain all changes of the previous PR until that is merged and this PR is rebased so this PR doesn't have ~2200 changes but ~1800.

### DIFF
--- a/dev-tool/src/list_net/convert.rs
+++ b/dev-tool/src/list_net/convert.rs
@@ -496,6 +496,7 @@ mod tests {
             doc_with_action(&history[7], UserAction::Click),
             doc_with_action(&history[8], UserAction::Miss),
             doc_with_action(&history[9], UserAction::Miss),
+            /* -- next seq of query results (unique by `(session_id, query_count)`) -- */
             doc_with_action(&history[10], UserAction::Miss),
         ];
 
--- a/dev-tool/src/list_net/convert.rs
+++ b/dev-tool/src/list_net/convert.rs
@@ -1,0 +1,515 @@
+#![cfg(not(tarpaulin))]
+
+use std::{
+    convert::TryInto,
+    ffi::OsStr,
+    fs,
+    path::{Path, PathBuf},
+};
+
+use anyhow::Error;
+use itertools::Itertools;
+use log::debug;
+use serde::Deserialize;
+use structopt::StructOpt;
+use uuid::Uuid;
+
+use xayn_ai::{
+    list_net_training_data_from_history,
+    DayOfWeek,
+    DocumentHistory,
+    DocumentId,
+    QueryId,
+    Relevance,
+    SessionId,
+    UserAction,
+    UserFeedback,
+};
+
+use super::data_source::InMemorySamples;
+use crate::exit_code::NO_ERROR;
+
+/// Converts different file formats.
+///
+/// This is mainly used to prepare a samples file
+/// based on a directory containing soundgarden user
+/// data-frame csv files.
+#[derive(Debug, StructOpt)]
+pub struct ConvertCmd {
+    /// Directory in which soundgarden saved user data-frames (one csv file per user).
+    #[structopt(short = "d", long)]
+    from_soundgarden: PathBuf,
+    /// File in which the samples should be stored, e.g. `data.samples`.
+    #[structopt(short = "o", long)]
+    to_samples: PathBuf,
+}
+
+impl ConvertCmd {
+    pub fn run(self) -> Result<i32, Error> {
+        let ConvertCmd {
+            from_soundgarden,
+            to_samples,
+        } = self;
+
+        let mut storage = InMemorySamples::default();
+
+        for entry in fs::read_dir(&from_soundgarden)? {
+            let entry = entry?;
+            let path = entry.path();
+            if path.extension() != Some(OsStr::new("csv")) {
+                continue;
+            }
+
+            debug!("Loading history for {:?}.", path.file_name().unwrap());
+            let history = load_history(path)?;
+            debug!("Processing history");
+            for (inputs, target_prob_dist) in list_net_training_data_from_history(&history)? {
+                storage.add_sample(inputs.view(), target_prob_dist.view())?;
+            }
+        }
+
+        debug!("Write output file.");
+        storage.serialize_into_file(to_samples)?;
+
+        Ok(NO_ERROR)
+    }
+}
+
+//FIXME[follow up PR]: Change ltr feature extraction tests to share code with
+//                     this implementation and then to use the normal soundgarden
+//                     user data-frames instead of specially pre-processed csv
+//                     files.
+//
+//FIXME maybe verify the structural integrity of the file, currently it's expected
+// to have the records partitioned by query as well as sorted both in each query
+// (first ranked record to last ranked record) and in between queries (oldest
+// query to newest query).
+//
+/// Loads a history from a soundgarden user data-frame csv file.
+///
+/// The [`UserAction`] values will be inferred from the data in the
+/// same way soundgarden does so.
+fn load_history(path: impl AsRef<Path>) -> Result<Vec<DocumentHistory>, Error> {
+    let mut reader = csv::Reader::from_path(path)?;
+
+    let mut history = Vec::new();
+    for (counter, record) in reader.deserialize().enumerate() {
+        let record: SoundgardenUserDfRecord = record?;
+        history.push(
+            record.into_document_history_with_bad_user_action(
+                counter
+                    .try_into()
+                    .expect("User with more then 2^32-1 records."),
+            ),
+        );
+    }
+
+    fix_user_actions_in_history(&mut history);
+
+    Ok(history)
+}
+
+/// Struct to load a soundgarden user data-frame csv file record into.
+#[derive(Deserialize)]
+struct SoundgardenUserDfRecord {
+    /// Session Id, is an incremental unsigned integer in soundgarden.
+    session_id: u64,
+    /// Query Id, same for all queries using the same parameters.
+    query_id: u64,
+    /// User Id, arbitrary unsigned integer.
+    ///
+    /// We treat all data as if it's from the same user, so we
+    /// don't really need it but we use it to infer an appropriate
+    /// deterministic document id.
+    user_id: u64,
+    /// Number of days since the start of the soundgarden history.
+    ///
+    /// Starting with `Tue == 1`.
+    ///
+    /// As such `Mon` is `0`, `Wed` is `2` etc.
+    ///
+    /// The number is not limited to `0..=6`, e.g. all of `1`,`8`,`15` are `Tue`
+    day: usize,
+    /// The words of the query as *comma* separated string.
+    ///
+    /// This are not actual words but string encoded word ids.
+    /// E.g. `"2142,53423"` but treating `2142` as a word is not
+    /// a problem for our use-case.
+    query_words: String,
+    /// The url of the document.
+    ///
+    /// Theoretically this is also an "arbitrary" unsigned integer id, but
+    /// practically it's simpler to just treat it as a string.
+    url: String,
+    /// The domain of the document.
+    ///
+    /// Theoretically this is also an "arbitrary" unsigned integer id, but
+    /// practically it's simpler to just treat it as a string.
+    domain: String,
+    /// The relevance of the document.
+    relevance: Relevance,
+    /// The position of the document, *starting with `1`*.
+    position: usize,
+    /// The index of the query in the session it belongs to.
+    query_counter: usize,
+}
+
+impl SoundgardenUserDfRecord {
+    /// Creates a [`DocumentHistory`] from this record.
+    ///
+    /// The `user_action` field *will* be incorrect and
+    /// needs to be fixed separately. This is the case
+    /// as the other documents need to be known to
+    /// infer the [`UserAction`] correctly.
+    fn into_document_history_with_bad_user_action(
+        self,
+        per_user_result_counter: u32,
+    ) -> DocumentHistory {
+        let Self {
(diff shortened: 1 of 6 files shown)


dev-tool/src/list_net/data_source.rs
```diff
@@ -0,0 +1,567 @@
...
+///
+/// The storage is reasonably compact as ~5.3GiB of soundgarden user
+/// data-frames will produce a less then 975MiB serialized `.samples` file.
+// While there are many ways to improve on this (e.g. memory-mapped I/O), they are not relevant for our
+// use-case for now.
+#[derive(Serialize, Deserialize, Default)]
+pub(crate) struct InMemorySamples {
+    /// Vector of concatenated `inputs` and `target_prob_dist`, this slightly improves memory size and
+    /// cache locality, we can derive the number of document from the vectors length as the vectors length
+    /// is `nr_document * INPUT_NR_FEATURES + nr_document*1`, i.e. `nr_documents * 51`.
```
finiteprods, line 254: can you explain how "this slightly improves memory size and cache locality"?

dev-tool/src/list_net/data_source.rs
```diff
@@ -0,0 +1,567 @@
...
+                nr_docs=inputs.shape()[0],
+                nr_feats=ListNet::INPUT_NR_FEATURES,
+                inputs_shape=inputs.shape(),
+                prob_dist_shape=target_prob_dist.shape(),
+            );
+        }
+        let mut data = Vec::with_capacity(inputs.len() + target_prob_dist.len());
+        extend_vec_with_ndarray(&mut data, inputs);
+        extend_vec_with_ndarray(&mut data, target_prob_dist);
+        self.data.push(data);
```
finiteprods, line 339: perhaps you could call `data` on l336 something else e.g. `datum` so that e.g.
```rust
self.data.push(datum);
```
looks a bit less confusing

dev-tool/src/list_net/data_source.rs
```diff
@@ -0,0 +1,567 @@
...
+        Ok(())
+    }
+}
+
+/// Extend a vec with the elements of given array in logical order.
```
finiteprods, line 344:
```suggestion
/// Extend a vec with the elements of the given array.
```
do you need to say "in logical order"? I'm not even sure what exactly that means here.

dev-tool/src/list_net/data_source.rs
```diff
@@ -0,0 +1,567 @@
...
+    }
+
+    #[test]
+    fn test_adding_bad_matrices_fails() {
+        let mut storage = InMemorySamples::default();
+        let inputs = Array::zeros((10, 48));
+        let target_prob_dist = Array::ones((10,));
+        storage
+            .add_sample(inputs.view(), target_prob_dist.view())
+            .unwrap_err();
```
finiteprods, line 415: can we inspect / assert on the error? similarly further down the test

finiteprods (commented):
looks good, not much to fault. included some doc improvements, questions and minor suggestions.

dev-tool/src/list_net/data_source.rs
rustonaut: The alternative would be `Vec<(Vec<f32>, Vec<f32>)>` (or `Vec<(Array2<f32>, Array1<f32>)>` which has a bit more overhead due to storing a `Vec<f32>` and a `Ix2`, not that it matters with the data sizes we did got, through it might have made a difference with the sizes I expected to have).

So the sample features and target probability distribution might be in very different parts of the heap and in turn cause more cache misses.

I.e. given a cache line size of 64 (through modern processors have 128 or double fetch) the samples cover 31.25 cache lines and the target prob distribution covers 0.625 cache lines. So by storing them "inline" we will use 32 cache lines while when storing them separate we most likely will use 33 cache lines. (In case of modern CPUs it's more like 15 vs. 16 cache fetches). The memory/storage size improvement s are negligible, I guess.

Through that comment above is actually more though then I put into it when implementing it. 

I just thought it would be good to keep things together in memory and expected much larger data sources then what we have in the end. Then after implementing it (without much time investment) I realized that it was a pre-mature optimization but thought that it also shouldn't hurt to keep it.

So if you insist I can remove it.

dev-tool/src/list_net/data_source.rs
```diff
@@ -0,0 +1,567 @@
...
+            .map_err(Into::into)
+    }
+
+    /// Deserialize a instance from given file.
+    #[allow(dead_code)] //FIXME is used by training (added in part 3 of this PR)
+    pub(crate) fn deserialize_from_file(file: impl AsRef<Path>) -> Result<Self, Error> {
+        Self::deserialize_from(BufReader::new(File::open(file)?))
+    }
+
+    /// Deserialize a instance from given reader, preferably pass in a buffered reader.
```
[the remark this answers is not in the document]
rustonaut: I changed it to  `, using a buffered reader is preferred.` if that is ok?

dev-tool/src/list_net/data_source.rs
rustonaut: It's terminology taken from `ndarray`.

The "logical order" is the order in which you think about an array where the "last" index varies fastest
when iterating e.g. storing a `Array<f32, Somthing2D>` as `[ [x0_y0, x0_y1, ...], [x1_y0, ....],...]`.

Storage order can differ, and this function make sure to store things in logical instead of storage order.

I still can remove the part about logical order as it's what you would normal expect.

dev-tool/src/list_net/data_source.rs
```diff
@@ -0,0 +1,567 @@
...
+    ///
+    /// For now this is a range. I.e. `0..nr_of_samples`. We might need
+    /// to change this in the future, but for now this is simpler.
+    fn data_ids(&self) -> Result<RangeTo<usize>, Self::Error>;
+
+    /// Loads a sample and returns a reference to it.
+    ///
+    /// This method is `&mut self` as it might change the internal state
+    /// (due to loading/caching) and we also want to make sure that there
+    /// are no more lend out samples, when `load_sample` or `load_batch` are
```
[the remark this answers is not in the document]
rustonaut: `load_sample` and `load_batch` return references to the underlying data.

When I wrote this trait I expected that there is a reasonable chance that I might have to add more complex implementations of `Storage` then a in-memory storage in the future. In which case you like would have lend
out references into a cache of some form, which will potentially flush out old samples when `load_samples`/`load_batch` are called. (Hence why they are called `load_*` instead of e.g. `get_*`.)

With the knowledge I have now, multiple weeks after writing this, I'm pretty sure that it's not needed and I'm considering to remove the `Storage` trait and just fixate `DataSource` to always use in-memory storage. (In a follow up PR, not in this PR).

rustonaut (commented):
Added review comments.

acrrd:
> For the 500_000 samples with are ~5.4GiB unpacked on disk this will produce a 875MiB file. Which if compiled in release mode will load in ~1s (or ~50s if not in release mode).
 
How much takes to load the data from the csv? And how much is that in comparison to the training time?

rustonaut:
> How much takes to load the data from the csv? And how much is that in comparison to the training time?

Depends on how long you train in which way with which parameters.

Training which is "just good enough" to show how well something works can be done
in <2min (if compiled with `--release`). A training we would use in production would
then add an additional 30min-1h or so just to get some increasingly smaller improvements.

The initial conversion code took ~30min to run on my system, I really don't
want to run this before every single training (through now it's more like 2m30s, so it's now
running a single time 2m30 and then 1s per training attempt instead of 2m30s per
training attempt, which already saved me a lot of time, especially if trying things out
without `--release` which makes thinks way slower).

Also we have to do the conversion anyway the only thing we could strip out is the
serialization from/to a `.samples` files but given that it's a bincode/serde serialization
that are like 9 lines of code for the serialization/deserialization methods and a bit
overhead for the CLI.  

So this is not really just about saving time, it's mostly work which needs to be done anyway.

Through I also thought we will have much more data where not splitting it out would have
been a much bigger problem.

acrrd:
> Depends on how long you train in which way with which parameters.
>
> [...]

thanks!

rustonaut:
(I just rebased it on master again, as master changed since the last rebase)

dev-tool/src/list_net/data_source.rs
```diff
@@ -0,0 +1,567 @@
...
+    training_data_order: DataLookupOrder,
+    /// A container with all ids of all evaluation samples, returning them in randomized order.
+    evaluation_data_order: DataLookupOrder,
+}
+
+impl<S> DataSource<S>
+where
+    S: Storage,
+{
+    /// Creates a new `DataSource` from given storage and evaluation split.
```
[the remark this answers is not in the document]
finiteprods: thanks for the detailed explanation. I was familiar with the idea, without knowing that is what it is called!

general remark re the PS: I personally think it's ok to use the terminology that's common in the ML domain, even if the use of "set" maybe isn't the strictest sense, as long as it is footnoted as explained here. calling it "data source" is fine as well, but is quite a generic term.

dev-tool/src/list_net/data_source.rs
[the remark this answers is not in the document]
finiteprods: re the follow-up comment, sounds good.

dev-tool/src/list_net/data_source.rs
```diff
@@ -0,0 +1,567 @@
...
+    ///
+    /// See [`Storage.load_batch()`] about why this is `&mut self`.
+    fn load_batch<'a>(&'a mut self, ids: &'_ [DataId]) -> Result<Vec<Sample<'a>>, Self::Error>;
+}
+
+/// For now samples id's are always incremental integers form `0..nr_samples`.
+///
+/// Given that this is just used to handle shuffling and similar it's unlikely to
+/// ever change, we do explicitly not care about any consistency of the id to
+/// sample mapping between trainings.
```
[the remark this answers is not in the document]
finiteprods: got it, thanks.
```diff
-/// For now samples ids are always incremental integers form `0..nr_samples`.
+/// For now samples ids are always incremental integers from `0..nr_samples`.
```
(or just delete the `from`)
```diff
-///     - Through we do care about a deterministic evaluation split, as such
+///     - Though we do care about a deterministic evaluation split, as such
```

dev-tool/src/list_net/data_source.rs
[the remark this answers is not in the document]
finiteprods: it's ok but really should be a new sentence: `. A buffered reader is preferred`.
or this should be clear enough as well: `Deserialize an instance from the given (preferably buffered) reader`.

dev-tool/src/list_net/data_source.rs
finiteprods: thanks, no need to remove it, just that I found it hard to understand the comment w/o knowing what the "alternative" was that you had in mind. perhaps you can add some of the above e.g. the first sentence into the comment but I'll leave that to your judgement.

dev-tool/src/list_net/data_source.rs
finiteprods: fair enough! i would suggest either removing it, or adding "in the `ndarray` sense" just for clarity.

dev-tool/src/list_net/data_source.rs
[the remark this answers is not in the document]
rustonaut: I already did it in a follow up PR by now.

dev-tool/src/list_net/convert.rs
```diff
@@ -490,10 +489,10 @@ mod tests {
...
             doc_with_action(&history[4], UserAction::Miss),
-            /* ---- */
+            /* -- next seq of query results (unique by `(session_id, query_count)`) -- */
             doc_with_action(&history[5], UserAction::Miss),
             doc_with_action(&history[6], UserAction::Miss),
-            /* ---- */
+            /* -- next seq of query results (unique by `(session_id, query_count)`) -- */
             doc_with_action(&history[7], UserAction::Click),
             doc_with_action(&history[8], UserAction::Miss),
             doc_with_action(&history[9], UserAction::Miss),
```
finiteprods, line 498: shouldn't there be a divider after this one as well?

dev-tool/src/list_net/data_source.rs
```diff
@@ -326,22 +333,26 @@ impl InMemorySamples {
...
+        self.data.push(datum);
         Ok(())
     }
 }
 
-/// Extend a vec with the elements of given array in logical order.
+/// Extend a vec with the elements of the given array.
+///
+/// Elements are added in logical order independent of storage order,
+/// i.e. elements are added as if `data` was continuos and in standard
```
finiteprods, line 354: *continuous

dev-tool/src/list_net/data_source.rs
rustonaut: I already did clarify it in the code.

finiteprods (approved):
some minor stuff otherwise all good.

dev-tool/src/list_net/data_source.rs
rustonaut: :+1: Hm, why isn't `continuos` highlighted by the spell checker :thinking: Now the word is flagged :shrug:

PR #165, opened 2021-08-04
ListNet training with emulating XaynNet: Ty 1932 part 3 add support for different initializer

rustonaut:
Added support to pick any of `he-normal`,`he-uniform`,`glorot-normal`, `glorot-uniform` as weight matrix initializer.

The soundgarden ListNet training did use the tensorflow default (`glorot`, I think `glorot-uniform`). I initially implemented
`he-normal` but then noticed that it yields slightly better training results compared to our currently used ListNet.

Evaluating the current ListNet on the same samples yields a KL-Divergence of 0.31966 (rounded, exact number can slightly differ due to f32 imprecision and parallel execution).

So I added all the common/relevant initializer and compared them by training them on the 500_000 user
data-frames derived samples with: epochs=10,batch-size=32,evaluation-split=0.2.

The result was:

| initializer |  1st. Evaluation KL-Divergence (rounded)  | 2nd | 3rd | 4th |
|-------------|----|--|--|--|
| `he-normal` | 0.31786 | 0.30676556 | 0.3067569 | 0.30653685 | 0.3073197 |
| `he-uniform` | 0.31781 | 0.30660623 | 0.30695534 | 0.30692312 | 0.3177781 |
| `glorot-normal` | 0.30677 | 0.30643806 | 0.30632862 | 0.31784847 | 0.3063926 |
| `glorot-uniform` | 0.30651 | 0.30693167 | 0.3179018 | 0.30665034 | 0.30651706 |

As it can be seen the results are rather "unstable". I had many cases where `he-normal` yielded noticeable better results then ` he-uniform` but when I did ran the benchmarks it did not. Worse in nearly all times `glorot` was clearly worse then `he`, except when I did ran the benchmarks for the table :sob: .The only difference is that in the previous tests I think I didn't ran it using `--release` and I think it might have been before parallelizing execution. :thinking: 

Also it did failed with producing a `NaN` loss, which is a expected rare failure case as long as we don't use either batch normalization or gradient clipping.

Furthermore given that it's basically always better then the current ListNet I'm wondering if it's due too having a evaluation split with the same ratio but different samples or additional samples?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -434,6 +434,7 @@ dependencies = [
  "displaydoc",
  "env_logger",
  "float-cmp",
+ "indicatif",
  "itertools",
  "log",
  "ndarray",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -439,6 +439,7 @@ dependencies = [
  "log",
  "ndarray",
  "rand",
+ "rayon",
  "serde",
  "serde_json",
  "structopt",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -429,9 +429,20 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "base64",
+ "bincode",
+ "csv",
+ "displaydoc",
+ "env_logger",
+ "float-cmp",
+ "itertools",
+ "log",
+ "ndarray",
+ "rand",
  "serde",
  "serde_json",
  "structopt",
+ "thiserror",
+ "uuid",
  "xayn-ai",
 ]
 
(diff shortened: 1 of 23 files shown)


acrrd:
> Added support to pick any of `he-normal`,`he-uniform`,`glorot-normal`, `glorot-uniform` as weight matrix initializer.
> 
> The soundgarden ListNet training did use the tensorflow default (`glorot`, I think `glorot-uniform`). I initially implemented
> `he-normal` but then noticed that it yields slightly better training results compared to our currently used ListNet.

Why we want this here and cannot be done on tensorflow and then just implement what gives the best result?

rustonaut:
> Why we want this here and cannot be done on tensorflow and then just implement what gives the best result?

Because we don't get the exact same results in tensorflow and rust (not because of a bug, but because of subtle differences in thinks like sample shuffeling), and it's pretty simple tbh. I mean they are nearly the same method using nearly the same test.

Through tbh. we don't need to include this into master as it seems to really don't matter what initializer we choose.

acrrd:
> Through tbh. we don't need to include this into master as it seems to really don't matter what initializer we choose.

That, to reduce the code that we have to maintain or I was thinking to move this training code to another repo after we finish to merge everything. wdyt?

rustonaut:
Closed due to the comments above.

PR #168, opened 2021-08-09
TY-1844 unify ai paths

janpetschexain:
**References**

- [TY-1844]

**Summary**

- add a path resolver for the data in rust, such that it runs from the workspace as well as from the individual crate directories, and to only have a single source of truth for the paths to the data (which is `out/assets.json` whenever feasible)
- use the asset data in rust if it is defined/available 
- update the rust tests, benchmarks and examples (i also moved the validation from bin to example to use dev-dependencies, also it didn't belong to rubert's essential binaries in the first place)
- use the asset data in dart and update the tests
- update the ci to generate the asset data before using it in the tests etc

**Open questions**

- it seems impossible to apply this to the rust wasm tests, because `include_bytes!` only accepts a string literal which must be known at compile time
- i have no idea about html and if/how to change the wasm html example accordingly

**Todo**

- rebase and adjust once #156 is merged

[TY-1844]: https://xainag.atlassian.net/browse/TY-1844


rustonaut:
> it seems impossible to apply this to the rust wasm tests, because include_bytes! only accepts a string literal which must be known at compile time

It should not be necessary as:

> The file is located relative to the current file (similarly to how modules are found). The provided path is interpreted in a platform-specific way at compile time.

And as such the "working directory" from which you execute the the tests execution and/or test compilation should not matter.

janpetschexain:
> And as such the "working directory" from which you execute the the tests execution and/or test compilation should not matter.

that's true. my main issue with the wasm tests is that they still depend on the concrete, versioned path, whereas all other paths now depend on a single source of truth. i updated the pr summary to reflect this.

bindings/dart/test/mobile/utils.dart
```diff
@@ -6,25 +6,13 @@ import 'package:xayn_ai_ffi_dart/src/common/data/history.dart'
...
-  return SetupData(<AssetType, String>{
-    AssetType.smbertVocab: smbertVocab,
-    AssetType.smbertModel: smbertModel,
-    AssetType.qambertVocab: qambertVocab,
-    AssetType.qambertModel: qambertModel,
-    AssetType.ltrModel: ltrModel,
+SetupData mkSetupData() {
+  return SetupData({
+    for (final asset in getAssets().entries)
+      asset.key: '../../data/' + asset.value.suffix
```
rustonaut, line 15:
```suggestion
      asset.key: '../../' + asset.value.path
```

Instead of `../../` we could try to detect the workspace root, through it's potentially not worth the amount or work it implies.

data/src/bench/matmul.rs
```diff
@@ -0,0 +1,20 @@
+use std::{io::Result, path::PathBuf};
+
+use crate::resolve_path;
+
+const ASSET: &str = "bench_matmul_v0000";
+
+/// Resolves the path to the matrix multiplication benchmark data.
+pub fn data() -> Result<PathBuf> {
```
rustonaut, line 8:
```suggestion
pub fn data_dir() -> Result<PathBuf> {
```

This is not returning `data` (or a path to a file with data) but a dir path.

data/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+};
+
+use serde::Deserialize;
+use serde_json::{from_reader, from_value, Value};
+
+/// Resolves the path to the requested data.
+fn resolve_path(path: &[impl AsRef<Path>]) -> Result<PathBuf> {
+    let manifest = var_os("CARGO_MANIFEST_DIR")
+        .ok_or_else(|| Error::new(ErrorKind::NotFound, "missing CARGO_MANIFEST_DIR"))?;
+    let manifest = PathBuf::from(manifest).canonicalize()?;
```
rustonaut, line 27: Why not just use `manifest.parent().ok_or_else(...).to_path_buf()`, and make it independent of the current working directory?

data/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+use std::{
+    collections::HashMap,
+    env::{current_dir, var_os},
+    fs::File,
+    io::{BufReader, Error, ErrorKind, Result},
+    path::{Path, PathBuf},
+};
+
+use serde::Deserialize;
+use serde_json::{from_reader, from_value, Value};
```
rustonaut, line 21: I personally prefer prefixing constructors with the type/mod/crate instead of importing them as this removes ambiguities.

E.g. I would use `serde_json::from_reader` and `serde_json::from_value` directly in the one place they appear in the code. If I feel `serde_json` is to long/repeats to often I maybe would reexport it as `json` and then use `json::from_reader`, `json::from_value`.

data/src/qambert.rs
```diff
@@ -0,0 +1,39 @@
...
+pub fn model() -> Result<PathBuf> {
+    resolve_asset("qambertModel")
+}
+
+/// Resolves the path to the quantized QAMBert model.
+pub fn model_quant() -> Result<PathBuf> {
+    Ok(resolve_asset("qambertModel")?
+        .parent()
+        .ok_or_else(|| Error::new(ErrorKind::NotFound, "missing asset 'qambert'"))?
+        .join("qambert-quant.onnx"))
```
rustonaut, line 23:
```suggestion
        .with_file_name("qambert-quant.onnx"))
```

data/src/smbert.rs
```diff
@@ -0,0 +1,39 @@
...
+pub fn model() -> Result<PathBuf> {
+    resolve_asset("smbertModel")
+}
+
+/// Resolves the path to the quantized SMBert model.
+pub fn model_quant() -> Result<PathBuf> {
+    Ok(resolve_asset("smbertModel")?
+        .parent()
+        .ok_or_else(|| Error::new(ErrorKind::NotFound, "missing asset 'smbert'"))?
+        .join("smbert-quant.onnx"))
```
rustonaut, line 23:
```suggestion
        .with_file_name("smbert-quant.onnx"))
```

data/src/test/ltr.rs
```diff
@@ -0,0 +1,30 @@
...
+
+const ASSET: &str = "ltr_test_data_v0000";
+
+/// Resolves the path to the LTR feature extraction test data.
+pub fn data() -> Result<PathBuf> {
+    resolve_path(&["data", ASSET, "feature_extraction"])
+}
+
+/// Resolves the path to the intermediate LTR test model.
+pub fn intermediate() -> Result<PathBuf> {
```
rustonaut, line 13:
```suggestion
pub fn training_intermediates() -> Result<PathBuf> {
```

data/src/test/ltr.rs
```diff
@@ -0,0 +1,30 @@
+use std::{io::Result, path::PathBuf};
+
+use crate::resolve_path;
+
+const ASSET: &str = "ltr_test_data_v0000";
+
+/// Resolves the path to the LTR feature extraction test data.
+pub fn data() -> Result<PathBuf> {
```
rustonaut, line 8:
```suggestion
pub fn feature_extraction_test_cases() -> Result<PathBuf> {
```

rubert/benches/mbert.rs
```diff
@@ -48,10 +49,10 @@ macro_rules! bench_tract {
 fn bench_onnx(
     manager: &mut Criterion,
     name: &str,
-    vocab: impl AsRef<Path>,
-    model: impl AsRef<Path>,
+    vocab: Result<impl AsRef<Path>>,
+    model: Result<impl AsRef<Path>>,
 ) {
-    let tokenizer = TokenizerBuilder::from_file(vocab)
+    let tokenizer = TokenizerBuilder::from_file(vocab.unwrap())
```
rustonaut, line 55: The model path resolving is test specific maybe we want to move the unwrap into the `fn vocab()`,`fn model()` functions (they then using `resolve***.unwrap()`).

But I'm not sure about this.

But we never do not unwrap it.

rustonaut (changes requested):

Robert-Steiner:
Nice idea to use the `assets.json` file 👍
With the changes of #156 it will make the reading of the assets a bit simpler. I didn't like the `assets.json` because in my option it was too hard to read. Therefore I replaced it with a [static json file](https://github.com/xaynetwork/xayn_ai/pull/156/files#diff-0bcfa4d5d47b43679a40f5a7bfd737f2de808280dcdc3fc02203c5127077632e). That means that we later can remove the `Install gomplate` and `Build asset artifacts` in the rust jobs.

> i have no idea about html and if/how to change the wasm html example accordingly

after #156 has been merged we could create a symlink of `assets_manifest.json` to the wasm example folder and read the data paths from the manifest. 

I will take a deeper look tomorrow

Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [workspace]
 members = [
+    "data",
```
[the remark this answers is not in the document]
janpetschexain: makes sense, i'll move it to a new `test-utils` crate.

data/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+
+#[derive(Deserialize)]
+struct Asset {
+    name: String,
+    path: String,
+}
+
+#[derive(Deserialize)]
+struct Assets {
+    assets: Vec<Value>,
```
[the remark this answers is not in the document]
janpetschexain: unfortunately this doesn't work because the maps in the array are no homogeneuous, for example the maps related to wasm will crash the deserialization. this is why i had to go the `Value` way.

bindings/dart/test/mobile/utils.dart
janpetschexain: the dart `Asset` doesn't have a `path` field but only a `suffix` field (in contrast to the `json` file), hence the suggestion won't work.

data/src/lib.rs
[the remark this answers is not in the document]
rustonaut: I think this is more like a bug of the templates.

The relevant part is e.g.:
```handlebars
      "name": "wasmModule",
      {{ if ( $wasm_module_path | file.Exists ) -}}
      "suffix": "{{ $wasm_module_suffix }}",
      "checksum": "{{ $wasm_module_path | file.Read | crypto.SHA256 }}",
      "path": "{{ $wasm_module_path }}"
      {{ else -}}
      "suffix": "",
      "checksum": ""
      {{ end -}}
 ```
 
 but if we anyway set `"checkum"` and `"suffix"` to `""` couldn't we just do the same for `"path"`?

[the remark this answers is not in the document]
Comment (Robert-Steiner) [reply data/src/lib.rs:54]: > but if we anyway set "checkum" and "suffix" to "" couldn't we just do the same for "path"?
yes that shouldn't be an issue

File: bindings/dart/test/mobile/utils.dart (hunk 1 above)
Comment (rustonaut) [reply bindings/dart/test/mobile/utils.dart:15]: This is a bit troublesome because the whole idea behind `path` is to represent the path relative to the workspace root to the different assets, while the suffix is mobile on mobile specific and could theoretically diverge from path.

I'm wondering if it could make sense to add `path` to `Asset` :thinking: 

I mean the overhead this causes if complete negligible.

Or we somehow for tests set `suffix` = `path`, or something similar??

Comment (rustonaut) [reply bindings/dart/test/mobile/utils.dart:15]: Or we go the other way around:

- remove `path` from `assets.json
- force tests to use a "data dir" where assets are placed in by their path, so
  that `{DATA_DIR}/{suffix}` is always valid. 
- depending on the context `DATA_DIR` could be computed (based on the manifest dir) or
  fixed.
  
That probably would be simpler and now encode that fact that path and suffix can not diverge.

Just as an alternative idea.

Comment (Robert-Steiner) [reply bindings/dart/test/mobile/utils.dart:15]: > I'm wondering if it could make sense to add path to Asset 

Not sure about this because we will only use it internally for the tests and it could accidentally be used (instead of the suffix)  by team blue.

The intent of `path` was to only use it in the release ci to find the assets.

File: data/src/lib.rs (hunk 4 above)
Comment (janpetschexain) [reply data/src/lib.rs:21]: i agree that this makes sense in a context with different constructors. in this case, those are the only constructors and there shouldn't be any danger of confusion. also hovering over the function shows the tooltip/library if needed for clarification.

File: rubert/benches/mbert.rs (hunk 9 above)
Comment (janpetschexain) [reply rubert/benches/mbert.rs:55]: although we never not unwrap them, i'd prefer the utils crate to provide non-panicking/error handling if downstream crates may want to use it.

File: data/src/lib.rs [hunk 12]
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/// Reads the asset paths.
+fn read_assets() -> Result<HashMap<String, PathBuf>> {
+    from_reader::<_, Assets>(BufReader::new(File::open(resolve_path(&[
+        "out",
+        "assets.json",
+    ])?)?))
```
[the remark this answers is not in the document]
janpetschexain: i'll change it into multiple lines.

data/src/lib.rs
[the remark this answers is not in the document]
janpetschexain: i agree, i'd suggest to fix this in a follow up pr.

rustonaut (approved):

data/src/lib.rs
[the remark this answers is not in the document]
janpetschexain: if we can simplify it to only have one of the two approaches, it probably makes sense.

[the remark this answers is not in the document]
Robert-Steiner: What is the benefit of removing `path`? Is the value of suffix then `<data_dir>/<suffix>`?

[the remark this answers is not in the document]
Robert-Steiner: I thought a bit more about the idea of removing `path`. The only issue I see here is that the wasm artifacts are not stored in the `data_dir` so `<data_dir>/<suffix>` wouldn't work in the CI. However we can remove `path` in the [static json file](https://github.com/xaynetwork/xayn_ai/pull/156/files#diff-0bcfa4d5d47b43679a40f5a7bfd737f2de808280dcdc3fc02203c5127077632e) so that we only have `suffix`. 
wdyt?

PR #176, opened 2021-08-13
TY-2015 PoC WASM threads

Robert-Steiner:
How to run it:

```shell
cd xayn-ai-ffi-wasm 
rustup default nightly
RUSTFLAGS='-C target-feature=+atomics,+bulk-memory,+mutable-globals' \
  wasm-pack build --target web --release --no-typescript --out-dir example/pkg \
  -- -Z build-std=panic_abort,std
cd example/pkg

# The wasm-opt version that is downloaded by wasm-pack does not support threads therefore we need to use a newer version (101). 
# You can download it here: https://github.com/WebAssembly/binaryen/releases/tag/version_101

~/Downloads/binaryen-version_101/bin/wasm-opt -Oz --detect-features xayn_ai_ffi_wasm_bg.wasm -o xayn_ai_ffi_wasm_bg.wasm

cd ..
npm install
webpack build ./index --mode production -o dist --output-filename index.js
python3 server.py
# open: http://localhost:8000/ in your browser
```


rustonaut:
As a side not `wasm-pack` will use any `wasm-opt` which is available in the `PATH`. 

So instead of disabling `wasm-opt` and manually running it we can make sure a up-to-date 
`wasm-opt` binary is in the path and then it will be used.

rustonaut:
Can this be closed? (To cleanup the PR list.)

Robert-Steiner:
> Can this be closed? (To cleanup the PR list.)

Sure :)

PR #178, opened 2021-08-16
Ty 1932 part 4 support chunking data source

rustonaut:
The data source currently has one sample per user but we treat it as many samples for one user.

For XaynNet training emulation we need to pretend each user has multiple samples (in PR part-6 each
user has multiple samples but we still need to merge some them as users in the training database have 
too little history compared to real users).

### DIFF
--- a/dev-tool/src/list_net/data_source.rs
+++ b/dev-tool/src/list_net/data_source.rs
@@ -5,7 +5,6 @@ use std::{
     error::Error as StdError,
     fs::File,
     io::{BufReader, BufWriter, Read, Write},
-    ops::RangeTo,
     path::Path,
     u64,
 };
@@ -13,6 +12,7 @@ use std::{
 use anyhow::{bail, Error};
 use bincode::Options;
 use displaydoc::Display;
+use itertools::Itertools;
 use log::debug;
 use ndarray::{Array1, Array2, ArrayBase, ArrayView, Data, Dimension};
 use rand::{prelude::SliceRandom, Rng};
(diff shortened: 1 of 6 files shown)


dev-tool/src/list_net/data_source.rs
```diff
@@ -36,65 +33,150 @@ where
...
     /// Creates a new `DataSource` from given storage and evaluation split.
     ///
     /// # Errors
     ///
-    /// - If `evaluation_split` is less than 0 or not a normal float.
-    /// - If there are no samples.
-    /// - If there are no training samples with the given `evaluation_split`.
-    /// - If there are no evaluation samples with the given `evaluation_split`,
+    /// - If `evaluation_split` is less than 0, greater then 1.0 or not a normal float.
+    /// - If there are no training samples with the given `evaluation_split`, which is not exactly 1.0.
```
Robert-Steiner, line 42:
```suggestion
    /// - If there are no training samples with the given `evaluation_split`, which are not exactly 1.0.
```

dev-tool/src/list_net/data_source.rs
```diff
@@ -36,65 +33,150 @@ where
...
+    /// - If there are no evaluation samples with the given `evaluation_split`, which is not exactly 0.0.
```
Robert-Steiner, line 43:
```suggestion
    /// - If there are no evaluation samples with the given `evaluation_split`, which are not exactly 0.0.
```

dev-tool/src/list_net/data_source.rs
```diff
@@ -273,9 +340,15 @@ pub struct InMemorySamples {
 pub enum StorageError {
     /// The database was parsed successfully but contains broken invariants at index {at_index}.
     BrokenInvariants { at_index: usize },
+
+    /// A out-of-bounds DataId (failed: {id} < {exclusive_id_upper_bound}) was used with the storage, this is most likely a bug.
```
Robert-Steiner, line 344:
```suggestion
    /// An out-of-bounds DataId (failed: {id} < {exclusive_id_upper_bound}) was used in the storage, this is most likely a bug.
```

dev-tool/src/list_net/data_source.rs
```diff
@@ -309,8 +382,36 @@ impl InMemorySamples {
         Ok(self_)
     }
 
+    /// Returns the number of samples in the storages.
+    ///
+    /// The samples can be referred to by usize ids from
+    /// `0` to `number_of_samples`.
```
Robert-Steiner, line 388: referred by their id?

Robert-Steiner (approved):
looks good 👍
Just found some typos

dev-tool/src/list_net/data_source.rs
```diff
@@ -36,65 +33,150 @@ where
...
+impl DataSource {
     /// Creates a new `DataSource` from given storage and evaluation split.
     ///
     /// # Errors
     ///
-    /// - If `evaluation_split` is less than 0 or not a normal float.
-    /// - If there are no samples.
-    /// - If there are no training samples with the given `evaluation_split`.
-    /// - If there are no evaluation samples with the given `evaluation_split`,
+    /// - If `evaluation_split` is less than 0, greater then 1.0 or not a normal float.
```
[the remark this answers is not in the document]
rustonaut: There are normal IEEE and denormal IEEE floats.

The list of denormal IEEE floats contain:

- NaN
- +-Inf
- subnormals
- 0 (through we accept 0)

I can update the doc to:

```
/// - If `evaluation_split` is less than 0, greater then 1.0 or either of NaN, Inf or subnormal.
``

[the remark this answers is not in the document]
Robert-Steiner: That would be nice! 👍

dev-tool/src/list_net/data_source.rs
rustonaut: It's not really their id i.e. the id is not a property of the sample.

It's just a mechanism to temporary refer to a sample without having the sample yet.

PR #180, opened 2021-08-17
TY 2034 Add parallel WASM initializer.

rustonaut:
Adds support for building parallelized WASM:

- [x] support both parallelized and non parallelized WASM
- [x] update Makefile.toml
- [x] make sure commands like `cargo check` or `cargo clippy` still run like before.
- [x] make sure the dart example works on chrome
- [x] make sure that besides the `Atomic.wait()` on main thread issue it seem to work on firefox
- [x] consider removing `xayn-ai-ffi-wasm/example/` there seem to be little value for this by now,
      it doesn't manual test the dart bindings (which is what matters) and it's a bunch of additional
      maintenance overhead.

- This does not update the CI, the CI will run like before, but this also mean it won't run or build the parallelized
WASM.

- This doesn't run wasm bindgen tests with parallelized wasm, it's not clear how to do so tbh.

- This doesn't include a bundler, at least for now we don't need it.

This should be merged after similar changes for the C-FFI had been made.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -100,20 +100,13 @@ env = { WASM_VERION="wasm_bindings", WASM_OUT_DIR_PATH="bindings/dart/example/as
 run_task = { name = [ "build-dart", "build-wasm", "generate_assets_metadata"] }
 
 [tasks.build-wasm]
-dependencies = ["build-wasm-parallel", "build-wasm-sequential"]
+env = { DISABLE_WASM_THREADS = { value = "0", condition = { env_not_set = ["DISABLE_WASM_THREADS"] } } }
+run_task = { name= ["build-wasm-parallel", "build-wasm-sequential"] }
+
 
 [tasks.build-wasm-parallel]
 env = { RUSTC_BOOTSTRAP=1 }
-condition_script = ["""
-  # We can't use env_false as it requires the variable to be set
-  # and trying to set a default value for it in `tasks.build-wasm`
-  # seems to not work well either.
-  if [ "${DISABLE_WASM_THREADS:-0}" = 0 ]; then
-    exit 0
-  else
-    exit 1
-  fi
-"""]
+condition = { env_false = ["DISABLE_WASM_THREADS"] }
 script = ["""
   if ! [ "$(wasm-opt --version | grep -oE "version [0-9]+" | grep -oE "[0-9]+" )" -ge 101 ]; then
       echo "Make sure wasm-opt is in the PATH and has version >= 101" >&2
@@ -126,7 +119,7 @@ script = ["""
       --no-typescript \
       --target web \
       --release \
-      --out-dir ../$WASM_OUT_DIR_PATH --out-name genesis \
+      --out-dir ../${WASM_OUT_DIR_PATH:?WASM_OUT_DIR_PATH needs to be set} --out-name genesis \
       -- \
       -Z build-std=panic_abort,std
   rm $WASM_OUT_DIR_PATH/.gitignore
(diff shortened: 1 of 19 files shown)


bindings/dart/lib/src/web/ffi/library.dart
```diff
@@ -16,9 +18,26 @@ external _Promise<Wasm> _init([
...
+
 /// Initializes the wasm module.
 ///
 /// If `moduleOrPath` is a `RequestInfo` or `URL`, makes a request and
 /// for everything else, calls `WebAssembly.instantiate` directly.
-Future<Wasm> init([dynamic moduleOrPath]) async =>
-    promiseToFuture(_init(moduleOrPath));
+Future<Wasm> init([dynamic moduleOrPath]) async {
+  final wasm = await promiseToFuture<Wasm>(_init(moduleOrPath));
+  if (_needsParallelInit()) {
```
acrrd, line 33: Can't we just have an `_initThreadPool` that does nothing in the wasm that does not support parallelization?

rustonaut: Yes I will change it.

xayn-ai-ffi-wasm/Cargo.toml
```diff
@@ -11,10 +11,21 @@ console_error_panic_hook = "0.1.6"
...
-xayn-ai = { path = "../xayn-ai" }
+wasm-bindgen = { version = "=0.2.76", features = ["serde-serialize"] }
 xayn-ai-ffi = { path = "../xayn-ai-ffi" }
 
+# We use the "atomics" `target_feature` to enable parallelism instead of a
+# crate feature. This is necessary, as using a "normal" feature will break
+# "cargo clippy --all-targets --all-features" and similar. Furthermore we
+# always want to use parallelism if our target supports it, which this
+# setups represents fairly well.
+[target.'cfg(all(target_arch="wasm32", target_feature = "atomics"))'.dependencies]
```
Robert-Steiner, line 22: `target.'cfg...` is the reason why we need to switch to `resolver = "2"` right?

Makefile.toml
```diff
@@ -82,14 +85,62 @@ script = ["""
...
+"""]
+script = ["""
+  if ! [ "$(wasm-opt --version | grep -oE "version [0-9]+" | grep -oE "[0-9]+" )" -ge 101 ]; then
+      echo "Make sure wasm-opt is in the PATH and has version >= 101" >&2
+      echo "Download Url: https://github.com/WebAssembly/binaryen/releases/tag/version_101" >&2
+      exit 1
+  fi
+  echo "build parallel wasm"
+  RUSTFLAGS='-C target-feature=+atomics,+bulk-memory,+mutable-globals' \
+    rustup run nightly \
```
janpetschexain, line 124: this fails to compile with the latest `nightly` version, one recent version which worked on my machine is `2021-08-17`. it shouldn't matter too much which version we choose, but we should pin it to a working version.
```suggestion
    rustup run nightly-2021-08-17 \
```

xayn-ai-ffi-wasm/Cargo.toml
```diff
@@ -12,9 +12,20 @@ getrandom = { version = "0.2.3", features = ["js"] }
...
+wasm-bindgen-rayon = {  version = "=1.0.3",  features = ["no-bundler"] }
+xayn-ai = { path = "../xayn-ai", features=["parallel"] }
```
janpetschexain, line 24:
```suggestion
xayn-ai = { path = "../xayn-ai", features = ["parallel"] }
```

janpetschexain (commented):
we should use a fixed nightly version for the parallel wasm compilation, other than that it should be good to go.

Makefile.toml
acrrd: Didn't we decide to use the bootstrap flag on the stable compiler? Otherwise we should use the nightly that correspond to the stable version we are using.

rustonaut: Right, I didn't write that down and I forgot over the 5 days skip.

I also haven tried it yet.

janpetschexain: if the flag works we should try it as well for #199, otherwise we should use `nightly-2021-03-25` as in #199 which corresponds to the nightly of the same date as the sable 1.51 release.

rustonaut: Seem to work :+1: 

I will push commits later today.

janpetschexain (approved):

PR #181, opened 2021-08-17
Fix bug in cargo make build-mobile.

rustonaut:
Specifically both `build-ios` and `build-android` tried to potentially build _all_ workspaces. Including `xayn-ai-ffi-wasm` and
`dev-tool`. But `dev-tool` has neither lib nor main entry point on mobile and `cargo ndk` still tries to forceful build a
binary. This then fail making `cargo make build-android` fail.

The `cargo lipo` command used on `cargo make build-ios` has conceptually the same problem but only considers workspaces which create `staticlib`s, which means the bug should not have triggered.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -314,7 +314,7 @@ jobs:
       - name: Build iOS library
         env:
           RUSTFLAGS: ${{ env.PRODUCTION_RUSTFLAGS }}
-        run: cargo lipo --targets ${{ matrix.target }} --release  -p xayn-ai-ffi-c
+        run: cargo lipo --targets ${{ matrix.target }} --release -p xayn-ai-ffi-c
 
       - name: Strip symbols
         run: |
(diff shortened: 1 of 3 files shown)


Robert-Steiner:
Nice! This will also close [TY-1880]. I just triggert the release ci to check whether we also need to update the cmd there.

Robert-Steiner:
Ok we also need to set the flag in the release ci android/ios jobs https://github.com/xaynetwork/xayn_ai/actions/runs/1142334821

.github/workflows/release.yaml
```diff
@@ -314,7 +314,7 @@ jobs:
       - name: Build iOS library
         env:
           RUSTFLAGS: ${{ env.PRODUCTION_RUSTFLAGS }}
-        run: cargo lipo --targets ${{ matrix.target }} --release
+        run: cargo lipo --targets ${{ matrix.target }} --release  -p xayn-ai-ffi-c
```
Robert-Steiner, line 317:
```suggestion
        run: cargo lipo --targets ${{ matrix.target }} --release -p xayn-ai-ffi-c
```

Robert-Steiner (approved):
release ci works :) thanks!

PR #182, opened 2021-08-18
TY 2036 add generator for example data.

rustonaut:
- Add generator for example data (to dev-tool)
- Also extend the number of documents in the default example data to 30.


xayn-ai/src/data/document.rs
```diff
@@ -15,7 +15,7 @@ use super::document_data::DocumentDataWithMab;
 
 #[repr(transparent)]
 #[cfg_attr(test, derive(Default))]
-#[derive(Debug, PartialEq, Eq, Clone, Hash, Serialize, Deserialize, Display)]
+#[derive(Debug, PartialEq, Eq, Clone, Copy, Hash, Serialize, Deserialize, Display)]
```
acrrd, line 18: Is there a reason to constraint this to `Copy`? It remove some clone but we are going to copy always 128bit instead of the smaller reference.

dev-tool/src/call_data/generate.rs
```diff
@@ -0,0 +1,492 @@
...
+    pub fn run(self) -> Result<i32, Error> {
+        let GenerateCallDataCmd {
+            out,
+            snippets,
+            number_of_current_documents,
+            number_of_historic_queries,
+            lengthen_urls,
+        } = self;
+
+        if !(MIN_QUERY_NR_DOCS..=MAX_QUERY_NR_DOCS).contains(&number_of_current_documents) {
```
acrrd, line 88: Why do we have this? I think we can have only a lower bound of 1 but if in the future we need 100 we can just input 100 and have 30 as default.

data/example_gen_snippets.json
```diff
@@ -0,0 +1,168 @@
+{
```
acrrd, line 1: Maybe rename to `documents_snippets`? This name make me thinks that this are example of generated snippets.

acrrd (commented):
Looks good but I'll prefer to remove `example3` and `example_long` an reduce `example` to 30 history and 30 documents.

xayn-ai/src/data/document.rs
rustonaut: It should still be faster to copy a 128bit value then to use a reference to a 128bit value, in general.

The few (64) additional bites copied on the stack normally do not out-weight the benefits from not
needing to de-reference the value when using it, even if it's mostly copied around.

There was a good article about when to pass by-value and when to pass by reference a few years ago,
let me see if I can find it.

If I remember correctly it was something like:

- On 64bit passing 128bit by value can be either faster or similar fast depending on circumstance.

- On 32bit passing 128bit by value might be slightly slower, but you need to try really hard to measure it as in most contexts it is such a small difference that it is drowned by noise.

I mean the only way passing a 128bit value is potentially slower (in this use-case) is if:

- You could pass it to function in the register instead of using the stack.
- and we pass it through a bunch of function calls, through register before using it.

E.g. if we have `foobar(&uuid)` and `foobar` is a leaf function we do pass the `&` 
most likely in registers (at least on 32bit arm EABI) but then we de-reference it 
(from the stack), making it maybe even (neglible) slower then just directly passing
it through the stack from the get to go. 

And on 64bit we can pass uuid directly through in registers (normally).

Sure things are more complex when multiple argument are involved etc.

But then the stack is (normally) in the L1 D-Cache so it's not that much slower.

And non of this code is in tight/hot loop.

Because of this I normally always pass UUIDs by Copy without thinking about it much,
I double we could measure a perf. difference between `&DocumentId` and `DocumentId` 
in our bussiness-use-case (reranking).

And `Copy` is often simpler to use.

dev-tool/src/call_data/generate.rs
rustonaut: Sure, I mainly used it to constraint the random generated history. I can change this to be 1+.

xayn-ai/src/data/document.rs
rustonaut: So should I undo the change? Or is it ok to keep it?

acrrd: Keep it

acrrd: Sorry I completely lost this comment.

acrrd:
Can we change the example document as said in my previous document?

rustonaut:
Sure, but why?

Especially example_long is useful for bench-marking as it as more realistic domain/url length.

acrrd:
Because we are not going to use it at the moment. We can generate them when we will need it.

rustonaut:
>  reduce example to 30 history and 30 documents.

I it's currently 20 historic query result in the history.
(Each having between 5-30 document.)

acrrd:
> > reduce example to 30 history and 30 documents.
> 
> I it's currently 20 historic query result in the history.
> (Each having between 5-30 document.)

I'll prefer to have something like before with one query, 30 documents and the same in the history so we can use those for the feedback loop.

rustonaut:
But this are complete unrealistic values which makes the while benchmark function kinda useless.

And I mean that's while there is/was `example3` which just contained one query with one historic query (which was the same query, for the feedback loop).

Not having any history (besides the current query) means that a lot of branches in the domain reranker are not hit at all.

rustonaut:
I mean the main problem is that our data-structure (call data, but also data in the C-FFI) is really bad (size/performance wise).

And even just a history of 19+1 search results is ~192KiB in JSON.

acrrd:
The purpose of this task was incrementing the example that we had before from 10 documents to 30 because that is the number of document that the app is passing us. At the moment we are mainly interested in see what speed up we have with the current changes to bert and multithread. 
In the future I would like to have a more comprehensive suite of benchmarks: for each system and also from dart  (with and without the need of a phone), at that point we can generate more realistic files.

rustonaut:
Rebased and examples where updated.

bindings/dart/example/assets/call_data/example.json
```diff
@@ -1,268 +1,788 @@
...
+    },
+    {
+      "id": "d1c41b13-435d-4e37-938a-9302438898dd",
+      "relevance": 0,
+      "user_feedback": 2,
+      "session": "32b42d89-ba38-47a0-8721-573a63402fef",
+      "query_count": 0,
+      "query_id": "e04dd090-2b57-4100-befa-4fa050560a06",
+      "query_words": "birds",
+      "day": 4,bindings/dart/example/assets/call_data/example.json
```
acrrd, line 166: this should not belong here :)

rustonaut: Wtf :scream_cat: 

Not running tests can really bite you, even when just regenerating some files :wink:

acrrd (approved):

PR #199, opened 2021-08-31
TY-2039 build wasm thread version in CI

Robert-Steiner:
**References**

- [TY-2039]

**Summary**

- split the building of wasm release artifacts in the release ci into a sequential and a parallel version
- the modification of `generate_asset_metadata.sh` and the adjustments of the dart code will be done in a follow up pr

[TY-2039]: https://xainag.atlassian.net/browse/TY-2039


janpetschexain:
the [release ci](https://github.com/xaynetwork/xayn_ai/actions/runs/1208626994) runs successfully

.github/workflows/release.yaml
```diff
@@ -331,23 +331,59 @@ jobs:
     needs: cargo-registry-cache
     runs-on: ubuntu-20.04
     timeout-minutes: 30
+    strategy:
+      matrix:
+        wasm_feature: [none, threads]
```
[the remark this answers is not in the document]
acrrd: changing `threads` to `parallel` yes. I'm not sure we want `wasm_threads` because in the future with simd we will probably want `none, threads, threads_simd`, I'm not sure if there are browser with simd but not shared buffer. otherwise will make sense to have `wasm_threads` and `wasm_simd`.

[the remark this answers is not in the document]
janpetschexain: according to the [wasm features table](https://webassembly.org/roadmap/) there are some cases where simd is available but threads are not. i guess it makes sense to split it into `wasm_threads: [sequential, parallel]` and later add a `wasm_simd: [single, multiple]` (or sth similar) which can then be combined in all possible configurations.

janpetschexain:
the [release ci](https://github.com/xaynetwork/xayn_ai/actions/runs/1209458381) runs successfully

.github/workflows/release.yaml
```diff
@@ -346,8 +358,12 @@ jobs:
           profile: minimal
           toolchain: ${{ env.RUST_STABLE }}
           target: wasm32-unknown-unknown
+          components: ${{ matrix.components }}
```
rustonaut, line 361: *Not sure*, but wouldn't it be better for caching to just  always fetch `stable` + `rust-src`?

.github/workflows/release.yaml
```diff
@@ -423,8 +448,8 @@ jobs:
         uses: ./.ci/build-asset-artifacts
         with:
           dart-ws: ${{ env.DART_WORKSPACE }}
-          wasm-version: ${{ needs.build-release-wasm-libs.outputs.version }}
-          wasm-out-dir-path: ${{ needs.build-release-wasm-libs.outputs.upload-path }}
+          wasm-version: ${{ needs.build-release-wasm-libs.outputs.sequential-version }}
```
rustonaut, line 451: Is it correct that this assumes the sequential version and the parallel version are the same?

 (Which is ok, but maybe add a comment).

.github/workflows/release.yaml
```diff
@@ -459,8 +484,12 @@ jobs:
       - name: Download WASM artifacts
         uses: actions/download-artifact@3be87be14a055c47b01d3bd88f8fe02320a9bb60 # v2.0.10
         with:
-          name: ${{ needs.build-release-wasm-libs.outputs.upload-name }}
-          path: ${{ needs.build-release-wasm-libs.outputs.upload-path }}
+          path: ${{ runner.temp }}/wasm_assets
```
rustonaut, line 487: Does this rely on `runner.temp` to not change?

.github/workflows/release.yaml
janpetschexain: i'm not sure either, i'll change it to always fetch the components, we can change it later on anyways.

.github/workflows/release.yaml
janpetschexain: this is going to be changed in the follow up pr #202. they are going to be different, but for now the asset generation has not yet been changed, hence the generated assets are the same.

.github/workflows/release.yaml
janpetschexain: the [github actions docs](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#runner-context) state that this temp dir is emptied at the beginning and end of each job, so i would assume that it stays populated between the steps within one job.

janpetschexain:
the [release ci](https://github.com/xaynetwork/xayn_ai/actions/runs/1213445657) runs successfully

rustonaut (approved):
lgtm

PR #202, opened 2021-09-06
TY-2069 Update assets script

Robert-Steiner:
**References**

- [TY-2069]
- [TY-2066]

**Summary**

- add an optional set of  `Feature`s to select the respective assets for sequential/parallel mobile/web in dart
- update the corresponding `getAssets()`, `SetupData()` functions and pick a default in the example app
- expand the asset generation script to generate sequential/parallel web assets depending on the input
- update the makefile to generate the sequential/parallel web assets depending on `DISABLE_WASM_THREADS`
- update the release ci to build the sequential/parallel web artifacts

**Todo**

- ~rebase once #207 is merged~

**Open questions**

- testing for mobile and sequential web was successful, but when trying to test the parallel web it loads indefinitely (i tried firefox) where i'm not sure if this is related to these changes or to the `run-web` script, to reproduce:
  - build with `DISABLE_WASM_THREADS=0 cargo make build-web`(or without explicitely setting the env var)
  - select `FeatureHint.webParallel` in the dart example's `main.dart` (l. 62)
  - run `cargo make run-web`, open in a browser and check the console

[TY-2069]: https://xainag.atlassian.net/browse/TY-2069
[TY-2066]: https://xainag.atlassian.net/browse/TY-2066

### DIFF
--- a/bindings/dart/example/lib/data_provider/data_provider.dart
+++ b/bindings/dart/example/lib/data_provider/data_provider.dart
@@ -1,7 +1,7 @@
-import 'package:xayn_ai_ffi_dart/package.dart' show SetupData;
+import 'package:xayn_ai_ffi_dart/package.dart' show FeatureHint, SetupData;
 
 /// Prepares and returns the data that is needed to init [`XaynAi`].
-Future<SetupData> getInputData() async =>
+Future<SetupData> getInputData([FeatureHint? hint]) async =>
     throw UnsupportedError('Unsupported platform.');
 
 String joinPaths(List<String> paths) {
--- a/bindings/dart/example/lib/data_provider/data_provider.dart
+++ b/bindings/dart/example/lib/data_provider/data_provider.dart
@@ -1,7 +1,7 @@
 import 'package:xayn_ai_ffi_dart/package.dart' show FeatureHint, SetupData;
 
 /// Prepares and returns the data that is needed to init [`XaynAi`].
-Future<SetupData> getInputData([FeatureHint? hint]) async =>
+Future<SetupData> getInputData(FeatureHint hint) async =>
     throw UnsupportedError('Unsupported platform.');
 
 String joinPaths(List<String> paths) {
--- a/bindings/dart/example/lib/data_provider/data_provider.dart
+++ b/bindings/dart/example/lib/data_provider/data_provider.dart
@@ -1,7 +1,7 @@
-import 'package:xayn_ai_ffi_dart/package.dart' show SetupData;
+import 'package:xayn_ai_ffi_dart/package.dart' show Feature, SetupData;
 
 /// Prepares and returns the data that is needed to init [`XaynAi`].
-Future<SetupData> getInputData({List<String> features = const []}) async =>
+Future<SetupData> getInputData({Set<Feature> features = const {}}) async =>
     throw UnsupportedError('Unsupported platform.');
 
 String joinPaths(List<String> paths) {
--- a/bindings/dart/example/lib/data_provider/data_provider.dart
+++ b/bindings/dart/example/lib/data_provider/data_provider.dart
@@ -1,7 +1,7 @@
-import 'package:xayn_ai_ffi_dart/package.dart' show FeatureHint, SetupData;
+import 'package:xayn_ai_ffi_dart/package.dart' show SetupData;
 
 /// Prepares and returns the data that is needed to init [`XaynAi`].
-Future<SetupData> getInputData(FeatureHint hint) async =>
+Future<SetupData> getInputData({List<String> features = const []}) async =>
     throw UnsupportedError('Unsupported platform.');
 
 String joinPaths(List<String> paths) {
--- a/bindings/dart/example/lib/main.dart
+++ b/bindings/dart/example/lib/main.dart
@@ -58,7 +58,7 @@ class _MyAppState extends State<MyApp> {
 
     // mobile is always loaded with multi-threading features, web only optionally
     debugPrint('start loading assets');
-    Logic.load(rootBundle, features: []).then((logic) {
+    Logic.load(rootBundle).then((logic) {
       debugPrint('loaded assets');
       // Calling `setState` after it was disposed is considered an error.
       if (mounted) {
--- a/bindings/dart/example/lib/main.dart
+++ b/bindings/dart/example/lib/main.dart
@@ -59,7 +59,8 @@ class _MyAppState extends State<MyApp> {
     super.initState();
 
     // mobile is always loaded with multi-threading features, web only optionally
-    final hint = kIsWeb ? FeatureHint.wasmSequential : null;
+    final hint =
+        kIsWeb ? FeatureHint.webSequential : FeatureHint.mobileParallel;
 
     debugPrint('start loading assets');
     Logic.load(rootBundle, hint).then((logic) {
--- a/bindings/dart/example/lib/main.dart
+++ b/bindings/dart/example/lib/main.dart
@@ -1,4 +1,3 @@
-import 'package:flutter/foundation.dart' show kIsWeb;
 import 'package:flutter/material.dart'
     show
         AppBar,
@@ -32,7 +31,6 @@ import 'package:flutter/material.dart'
 import 'package:flutter/services.dart' show rootBundle;
 import 'package:stats/stats.dart' show Stats;
 
-import 'package:xayn_ai_ffi_dart/package.dart' show FeatureHint;
 import 'package:xayn_ai_ffi_dart_example/debug/print.dart'
     if (dart.library.io) 'package:xayn_ai_ffi_dart_example/debug/mobile/print.dart'
     show debugPrintLongText;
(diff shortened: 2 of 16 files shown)


rsaffi:
You will need to rebase due to changes on `.github/workflows/release.yaml` in case #207 is merged before this PR.

janpetschexain:
> You will need to rebase due to changes on `.github/workflows/release.yaml` in case #207 is merged before this PR.

it probably makes sense that you merge first, this needs to be rebased once #199 is merged anyways.

janpetschexain:
the [release ci](https://github.com/xaynetwork/xayn_ai/actions/runs/1216877110) runs successfully

janpetschexain:
rebased, the [release ci](https://github.com/xaynetwork/xayn_ai/actions/runs/1217841395) runs successfully (fix for #207 already included)

janpetschexain:
the [release ci](https://github.com/xaynetwork/xayn_ai/actions/runs/1220825801) runs successfully

bindings/dart/example/lib/data_provider/data_provider.dart
```diff
@@ -1,7 +1,7 @@
-import 'package:xayn_ai_ffi_dart/package.dart' show FeatureHint, SetupData;
+import 'package:xayn_ai_ffi_dart/package.dart' show SetupData;
 
 /// Prepares and returns the data that is needed to init [`XaynAi`].
-Future<SetupData> getInputData(FeatureHint hint) async =>
+Future<SetupData> getInputData({List<String> features = const []}) async =>
```
acrrd, line 4: I think `FeatureHint` can stay, but instead of containing an enum of every possible combinations it only contains the features that can be enabled. We can have `FeatureHint { thread }`, and handle all combinations inside `getInputData`, in the future we have to only add `simd` to the enum.

acrrd: Also can use a `Set` instead of a `List`.
What I meant in the comment of the other PR was only to avoid one enum with a variant for every possible combinations because then we need to handle exponential explosion in the enum, in blue, and our code. With a set of hint we need to handle all the cases in only one place.

bindings/dart/example/lib/main.dart
```diff
@@ -59,11 +57,8 @@ class _MyAppState extends State<MyApp> {
     super.initState();
 
     // mobile is always loaded with multi-threading features, web only optionally
-    final hint =
-        kIsWeb ? FeatureHint.webSequential : FeatureHint.mobileParallel;
-
     debugPrint('start loading assets');
-    Logic.load(rootBundle, hint).then((logic) {
+    Logic.load(rootBundle, features: []).then((logic) {
```
acrrd, line 61: Since `[]` is the default we can omit it, or there is a reason why you want to make this explicit?

bindings/dart/lib/src/web/reranker/data_provider.dart
```diff
@@ -4,8 +4,20 @@ import 'package:xayn_ai_ffi_dart/src/common/reranker/data_provider.dart'
...
+  final redundantAssets = features.contains('webParallel')
+      ? [
+          common.AssetType.wasmSequentialModule,
+          common.AssetType.wasmSequentialScript,
+        ]
+      : [
+          common.AssetType.wasmParallelModule,
+          common.AssetType.wasmParallelScript,
+          common.AssetType.wasmParallelSnippet,
+        ];
```
rustonaut, line 18: There was the idea to have a uniform naming/interface returned from `getAssets` independent of parallel/sequential.

I.e. for either of `wasmSequentialModule` or `wasmParallelModule` get mapped to `wasmModule`, same for `wasm...Script` => `wasmScript` and `wasm...Snippet` to `wasmSnippet`(1).

(Then you also don't need to pass the `features` to `SetupData`)

(1): It's not given in the sequential version but that might change, so basically "if there is a `wasmSnippet` is should be loaded.  (We might have one or multiple assets on sequential, too. In general we shouldn't consider what we might have if we can change it later, but if we do map anything from `xxxParallellYYY` to `xxxYYY` then for consistency why not do so for `Snippet` too :shrug: (or not if you don't want to).

bindings/dart/lib/src/web/reranker/data_provider.dart
```diff
@@ -4,8 +4,20 @@ import 'package:xayn_ai_ffi_dart/src/common/reranker/data_provider.dart'
     as common show Asset, AssetType, baseAssets, SetupData;
 
 /// Returns a map of all assets required for initializing [`XaynAi`].
-Map<common.AssetType, common.Asset> getAssets() {
-  return common.baseAssets;
+Map<common.AssetType, common.Asset> getAssets(
+    {List<String> features = const []}) {
+  final redundantAssets = features.contains('webParallel')
```
rustonaut, line 9: 2 minor nit pickings which you can totally ignore if you want :wink: 

1. I personally prefer `webThreads` there are other forms of parallelism we could have implemented (but which would be x10-x100 more work to implement with **much** less speedups, etc. So it's not really an option, but because of this maybe going with `webThreads ` is better, but it doesn't really matter I guess)

2. If you feel fancy you could change `List<String>` to `Set<FeatureHint>` with `enum FeatureHint { webThreads, ...}` but it doesn't really matter, tbh.

generate_assets_metadata.sh
```diff
@@ -101,43 +101,63 @@ gen_data_assets_metadata() {
...
+
+gen_wasm_assets_metadata() {
+    local WASM_FEATURE=$1
+    local WASM_VERSION=$2
+    local WASM_OUT_DIR_PATH=$3
 
-        local UPDATED_ASSET=$(echo $ASSET | jq -c '. |= .+ {"fragments": []}' | jq -c 'del(.filename)')
+    local WASM_JS_NAME="genesis.js"
+    local WASM_MODULE_NAME="genesis_bg.wasm"
+    local WASM_SNIPPET_NAME="snippets/wasm-bindgen-rayon-7afa899f36665473/src/workerHelpers.no-bundler.js"
```
rustonaut, line 134: It would be nice if we could have a wildcard/glob lookup here which replaces the `hash` of the create. 

It's a bit prone to accidentally break when updating rayon.

rustonaut:
> testing for mobile and sequential web was successful, but when trying to test the parallel web it loads indefinitely (i tried firefox) where i'm not sure if this is related to these changes or to the run-web script, to reproduce: 

That is because this `parallel` build doesn't support Firefox (and future Chrome versions, if they fix their bug).

A follow up story is needed which runs the entry point in a worker, without using wasm "threads" to archive that.

bindings/dart/example/lib/data_provider/data_provider.dart
janpetschexain: i see, i'll change it.

bindings/dart/example/lib/main.dart
janpetschexain: it's more or less a leftover from testing. i'll remove it for now. once the web parallel feature is stabilized, we can set it as the default in the example app.

bindings/dart/lib/src/web/reranker/data_provider.dart
acrrd: we decided to user `parallel`to be consistent with the naming in rust.

janpetschexain: > I personally prefer webThreads there are other forms of parallelism we could have implemented (but which would be x10-x100 more work to implement with much less speedups, etc. So it's not really an option, but because of this maybe going with webThreads is better, but it doesn't really matter I guess)

i thought of it more like `webParallel` subsumes multiple things, for now only multi-threading, and later for example also simd. if we ever need a more fine-grained feature selection we can split it, but it's probably not needed.

> If you feel fancy you could change List<String> to Set<FeatureHint> with enum FeatureHint { webThreads, ...} but it doesn't really matter, tbh.

i'm going to change this as @acrrd requested it as well.

bindings/dart/lib/src/web/reranker/data_provider.dart
acrrd: We can do this in another PR so we can close this.

rustonaut: ok

janpetschexain: i agree, it's also not straight forward due to how and which variants of the `AssetType` enum are generated.

generate_assets_metadata.sh
janpetschexain: tbh i don't know how to achieve this: if we compile the crate with different versions of `wasm_bindgen_rayon`, then how is the script supposed to lockup/pick the correct version? i also don't know where this hash comes from, it's neither in the `Cargo.lock` nor one of the github commit hashes. for now i'm going to pin the version of `wasm_bindgen_rayon` in the `Cargo.toml` and make a note there to avoid breaking this accidentally.

rustonaut: Ok :+1: 

---

I guess you could do something like applying a bash/sh glob  on `${WASM_OUT_DIR_PATH}/${WASM_VERSION}/wasm-bindgen-rayon-*/...`. But if that glob covers multiple files (can it due to caching??) then this wouldn't work.

I.e. something like `local FOO; for path in glob/*; do FOO="${path}"; break; done`. Through getting the quoting right is a bit tricky as the glob (`*`) must not be in the quoting but you would like variable expansions to be in the quoting..

acrrd (approved):

rustonaut (approved):
lgtm

PR #205, opened 2021-09-07
Don't push tags to release repository 

acrrd:
This needs to be backported to let the release CI work correctly. At the moment it fails because the tags are already on the release repo and they cannot be overridden.


rustonaut:
Why are the tags already on the release repo, how did they end up there?

Is it because it retries to push all already previously pushed tags?

acrrd:
> Why are the tags already on the release repo, how did they end up there?
> 
> Is it because it retries to push all already previously pushed tags?

We created them. Yes it tries to push all the tags, we already removed this on master because the pushed tags were referring to a commit in this repo and not the corresponding commit in the release repo, so they were broken.

rustonaut (approved):
lgtm

PR #207, opened 2021-09-08
DEVOPS-485 Adjust CI for publishing of AI Assets to Staging or Production buckets

rsaffi:
~I will still create the relevant secrets, but as far as the changes go, this covers it.~ New secrets are there! 🎉 
The diff looks odd because it matches part of what was in the middle of the KeyCDN part, but actually all KeyCDN content is gone and now we just have a split logic for uploading assets to either Staging or Production S3 buckets.

Signed-off-by: Ricardo Saffi Marques <ricardo.marques@xayn.com>

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -466,102 +466,38 @@ jobs:
         run: sh download_data.sh
 
       - name: Install s3cmd
-        if: github.ref != 'refs/heads/release'
         uses: ./.ci/install-s3cmd
         with:
-          access-key: ${{ secrets.STAGING_ASSETS_BUCKET_ACCESS_KEY }}
-          secret-key: ${{ secrets.STAGING_ASSETS_BUCKET_SECRET_KEY }}
+          access-key: ${{ secrets.AI_ASSETS_BUCKET_ACCESS_KEY }}
+          secret-key: ${{ secrets.AI_ASSETS_BUCKET_SECRET_KEY }}
 
       - name: Upload assets to S3
-        if: github.ref != 'refs/heads/release'
+        id: s3-upload
         shell: bash
-        env:
-          BUCKET_URL: s3://xayn_ai_staging_assets
         run: |
           set -e
+
+          if [[ "${{ github.ref }}" = "refs/heads/release" ]]
+            then
+              export BUCKET_URL="s3://${{ secrets.AI_ASSETS_BUCKET_PROD }}"
+          else
+              export BUCKET_URL="s3://${{ secrets.AI_ASSETS_BUCKET_STG }}"
+          fi
 
           for ASSET in $(cat ${{ needs.build-asset-artifacts.outputs.json-metadata }} | jq -c '.assets[]'); do
               FRAGMENTS_LEN=$(echo $ASSET | jq -r '.fragments | length')
               if [ "$FRAGMENTS_LEN" = "0" ]; then
                   ASSET_URL_SUFFIX=$(echo $ASSET | jq -r '.url_suffix')
                   ASSET_PATH=$(echo $ASSET | jq -r '.path')
-                  s3cmd sync --acl-public --guess-mime-type $ASSET_PATH $ASSET_PATH ${{ env.BUCKET_URL }}/$ASSET_URL_SUFFIX
+                  s3cmd sync --acl-public --guess-mime-type $ASSET_PATH $ASSET_PATH ${BUCKET_URL}/$ASSET_URL_SUFFIX
               else
                   for FRAGMENT in $(echo $ASSET | jq -c '.fragments[]'); do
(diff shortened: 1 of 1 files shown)


rsaffi:
New secrets were created accordingly and are saved in 1Password.

Secrets that need to be deleted once this is merged:
* `KEYCDN_API_KEY`
  * Reason: KeyCDN deployment logic deprecated
* `KEYCDN_RSYNC_PRIVKEY`
  * Reason: KeyCDN deployment logic deprecated
* `KEYCDN_RSYNC_USERNAME`
  * Reason: KeyCDN deployment logic deprecated
* `KEYCDN_RSYNC_ZONE`
  * Reason: KeyCDN deployment logic deprecated
* `STAGING_ASSETS_BUCKET_ACCESS_KEY`
  * Reason: deprecated in favor of `AI_ASSETS_STG_BUCKET_ACCESS_KEY`
* `STAGING_ASSETS_BUCKET_SECRET_KEY`
  * Reason: deprecated in favor of `AI_ASSETS_STG_BUCKET_SECRET_KEY`

abargiela (approved):

.github/workflows/release.yaml
```diff
@@ -465,104 +465,48 @@ jobs:
...
-        if: github.ref != 'refs/heads/release'
+        id: s3-upload
         shell: bash
-        env:
-          BUCKET_URL: s3://xayn_ai_staging_assets
         run: |
           set -e
 
+          if [[ "${{ github.ref }}" = "refs/heads/release" ]]
+            then
```
janpetschexain, line 489:
```suggestion
          then
```

.github/workflows/release.yaml
```diff
@@ -465,104 +465,48 @@ jobs:
...
+              export BUCKET_URL="s3://${{ secrets.AI_ASSETS_BUCKET_PROD }}"
+          else
+              export BUCKET_URL="s3://${{ secrets.AI_ASSETS_BUCKET_STG }}"
```
janpetschexain, line 492: since the bucket keys have the word order `STG/PROD_BUCKET`, should we follow the same naming for the bucket url?
```suggestion
              export BUCKET_URL="s3://${{ secrets.AI_ASSETS_PROD_BUCKET }}"
          else
              export BUCKET_URL="s3://${{ secrets.AI_ASSETS_STG_BUCKET }}"
```

.github/workflows/release.yaml
rsaffi: 🤔 I've always indented if-clauses in bash like that:
```bash
if condition
  then
    (...)
  else
    (...)
fi
```

.github/workflows/release.yaml
rsaffi: Already updated the secrets and 1Password as well! 🚀

.github/workflows/release.yaml
janpetschexain: ah ok, not sure if there is any convention, i'm fine with this as well, but then the `else` needs to be indented.

janpetschexain (approved):

.github/workflows/release.yaml
rsaffi: As it is already, right? Or you think it should be indented one step further still?

rustonaut: I don't think there is a convention.

What I see most often is:

```
if condition ; then
   code
else
   code
fi
```

Note that the `;` is very important :wink: 

But IMHO it doesn't matter either way.

rsaffi: Yeah, I'll stick with what I went with as I think it gives some good visibility, if it's really ok. Let me merge this and get it out of your way, guys! :D

.github/workflows/release.yaml
```diff
@@ -465,104 +465,48 @@ jobs:
...
         if: github.ref != 'refs/heads/release'
         uses: ./.ci/install-s3cmd
         with:
-          access-key: ${{ secrets.STAGING_ASSETS_BUCKET_ACCESS_KEY }}
-          secret-key: ${{ secrets.STAGING_ASSETS_BUCKET_SECRET_KEY }}
+          access-key: ${{ secrets.AI_ASSETS_STG_BUCKET_ACCESS_KEY }}
+          secret-key: ${{ secrets.AI_ASSETS_STG_BUCKET_SECRET_KEY }}
+
+      - name: Install s3cmd (production)
+        if: github.ref = 'refs/heads/release'
```
janpetschexain, line 476: just a reminder to all of us, we should actually run stuff before we merge it to see [if it really works](https://github.com/xaynetwork/xayn_ai/actions/runs/1217825888/workflow#L476) :/
```suggestion
        if: github.ref == 'refs/heads/release'
```
it's going to be fixed in the next pr #202

PR #209, opened 2021-09-09
TY-2068 limit max threads

janpetschexain:
**References**

- [TY-2068]

**Summary**

- introduce a `maxNumberOfThreads` constant to limit the maximum number of threads to be used, it's currently set to 16 which is somewhat arbitrary

[TY-2068]: https://xainag.atlassian.net/browse/TY-2068

### DIFF
--- a/bindings/dart/lib/src/common/reranker/ai.dart
+++ b/bindings/dart/lib/src/common/reranker/ai.dart
@@ -1,3 +1,4 @@
+import 'dart:math' show min;
 import 'dart:typed_data' show Uint8List;
 
 import 'package:json_annotation/json_annotation.dart' show JsonValue;
@@ -76,15 +77,14 @@ class XaynAi {
   void free() => throw UnsupportedError('Unsupported platform.');
 }
 
+/// Maximum number of threads to be used for parallel features.
+const int maxNumberOfThreads = 16;
+
 /// Selects the number of threads used by the [`XaynAi`] thread pool.
 ///
 /// On a single core system the thread pool consists of only one thread.
 /// On a multicore system the thread pool consists of
-/// (the number of logical cores - 1) threads.
-int selectThreadPoolSize(int numberOfProcessors) {
-  if (numberOfProcessors > 1) {
-    return numberOfProcessors - 1;
-  } else {
-    return numberOfProcessors;
-  }
-}
+/// (the number of logical cores - 1) threads, but at most [`maxNumberOfThreads`] threads.
+int selectThreadPoolSize(int numberOfProcessors) => (numberOfProcessors > 1)
+    ? min(numberOfProcessors - 1, maxNumberOfThreads)
+    : 1;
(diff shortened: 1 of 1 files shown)


rustonaut:
Couldn't we just have done following?

```
int selectThreadPoolSize(int numberOfProcessors) {
  if (numberOfProcessors > 1) {
    return min(numberOfProcessors - 1, 32);
  } else {
    return numberOfProcessors;
  }
}
```

acrrd:
> Couldn't we just have done following one line change?
> ...

I was also about to ask why we are not doing it in dart.
If we realize that we are spawning too many threads in tests, ci, or in some tools we should be able to limit that without touching the ffi code.

janpetschexain:
of course we can do that, i'll change it accordingly.

since you guys already had an idea/opinion on where to add this, why didn't you say so in the first place (e.g. in the ticket or the talks)? i didn't know about this part of the dart code and it's quite annoying to learn about it after i've done the work. it would've saved time for all of us.

acrrd:
> since you guys already had an idea/opinion on where to add this, why didn't you say so in the first place (e.g. in the ticket or the talks)? i didn't know about this part of the dart code and it's quite annoying to learn about it after i've done the work. it would've saved time for all of us.

Yep, we should have mentioned that at least in the ticket. Sorry about that.

bindings/dart/lib/src/common/reranker/ai.dart
```diff
@@ -76,15 +77,14 @@ class XaynAi {
...
-/// (the number of logical cores - 1) threads.
-int selectThreadPoolSize(int numberOfProcessors) {
-  if (numberOfProcessors > 1) {
-    return numberOfProcessors - 1;
-  } else {
-    return numberOfProcessors;
-  }
-}
+/// (the number of logical cores - 1) threads, but at most [`maxNumberOfThreads`] threads.
+int selectThreadPoolSize(int numberOfProcessors) => (numberOfProcessors > 1)
```
acrrd, line 88: Don't know if it will be still good for readability, but we can simplify it further as:
`min(max(numberOfProcessors - 1, 1), maxNumberOfThreads)`.

bindings/dart/lib/src/common/reranker/ai.dart
```diff
@@ -76,15 +77,14 @@ class XaynAi {
...
 /// On a multicore system the thread pool consists of
-/// (the number of logical cores - 1) threads.
-int selectThreadPoolSize(int numberOfProcessors) {
-  if (numberOfProcessors > 1) {
-    return numberOfProcessors - 1;
-  } else {
-    return numberOfProcessors;
-  }
-}
+/// (the number of logical cores - 1) threads, but at most [`maxNumberOfThreads`] threads.
```
acrrd, line 87: and at least one.

acrrd (approved):

bindings/dart/lib/src/common/reranker/ai.dart
janpetschexain: the sentence refers to multi-core systems, hence i'd say the `(the number of logical cores - 1) threads` already implies at least one thread. i can still add it for explicitness.

acrrd: ok, with the diff I missed that was connected with the previous comment.

rustonaut:
> Yep, we should have mentioned that at least in the ticket. Sorry about that.

I second that from my side, I guess I did overlook some aspect on the handover. Sorry for this.

PR #221, opened 2021-09-20
Update docs and makefile toml

rustonaut:


Makefile.toml
```diff
@@ -93,7 +93,7 @@ script = ["""
   fi
 """]
 
-[tasks.run-web]
+[tasks.serve-web]
 script = ["""
```
acrrd, line 97: We can remove the `echo warning` now.

README.md
```diff
@@ -169,7 +173,7 @@ python3 -m http.server
 -   path: '../'
 +   git:
 +       url: git@github.com:xaynetwork/xayn_ai_release.git
-+       ref: master
++       ref: staging
```
acrrd, line 176: If you want to change this put `<branch>`

acrrd (commented):
Don't we need to update also the README for instruction to change the html?
Since now we cannot run the threaded version even in chrome we should change the default to sequential also in the `Makefile.toml`.

acrrd:
In the README we also need to change `_baseAssetUrl` from
`https://xayn_ai_staging_assets.s3-de-central.profitbricks.com` to `https://ai-assets.xaynet.dev`.

README.md
```diff
@@ -180,6 +184,12 @@ python3 -m http.server
 +   'https://xayn_ai_staging_assets.s3-de-central.profitbricks.com';
 ```
 
+Then use `cargo make build-web` and `cargo make serve-web` like above.
+
+This only works if the the release repos branch and this branch are in sync
```
acrrd, line 189: That the interface must mach is usual stuff and for example `flutter build web` gives error if the api we refer to is broken.
Are any case where the mismatch is not visibile at compile time and gives an error at runtime?

rustonaut: There could be not all breaking changes are always visible in the API.

acrrd: But is not different from any other dependency that we have. We can just add that if there is an unexpected runtime error to first check that the used version is compatible with the example.

acrrd (approved):

PR #222, opened 2021-09-21
TY-2075 refactor getAssets

Robert-Steiner:
Ticket:

- [TY-2075], [TY-2077], [TY-2091]

**Summary:**

- The `wasmModule` and` wasmScript` `AssetTypes` have been added and the` wasm<Feature>Module`, `wasm<Feature>Script` and `wasm<Feature>Snippet` have been removed.
- We can remove `wasm<Feature>Snippet` because `genesis.js` is taking care of loading it.
- Replaced the `wasmParallel` variant with the wasm features. This way team blue doesn't need to know what wasm features are necessary to choose the right wasm module. They just pass us the set of features that are available.
- ~Moved the `Feature` enum to the web package. Added a `Feature` class with a private constructor for the mobile package.~
	- ~The only issue with that solutions is that we need to disable some lints since `flutter analyze` does not respect conditional imports but uses the default import.~ (see https://github.com/xaynetwork/xayn_ai/pull/222#issuecomment-926756405)
- Removed the hardcoded snippets path. If a `snippets` folder exists, we just upload the whole folder.
- Split the `assets` array of `assets_metadata.json` into three arrays
	- `assets`  which contains the metadata of the data assets like vocab, model etc.
	- `wasm_assets`  which contains the metadata of the wasm assets like module and script
	- and `upload` which contains the `path` and `url_suffix` for each assets that need so be uploaded
	- this also simplified the code of uploading the assets in the release ci
- Removed the pinned version of `wasm-bindgen-rayon`.
- Updated `CHANGELOG.md`.

Ideas: 

- ~What about moving the `wasmSequential`  and `wasmParallel` hash map from the common package to the web package?~ (implemented)

---

`assets.dart`

```dart
/// Warning, this file is autogenerated. Don't modify this manually.
part of 'data_provider.dart';

/// Base assets that are required for both mobile and web.
///
/// The checksum is the sha256 hash of the asset.
/// To calculate the checksum run 'shasum -a 256 vocab.txt'.
final baseAssets = <AssetType, Asset>{
  AssetType.smbertVocab: Asset(
      'smbert_v0000/vocab.txt',
      Checksum(
          'ffb2398810fa977ee39e8a0103b07f9c2614f1f9825df53a2c99e71569e3451c'),
      []),
  AssetType.smbertModel: Asset(
      'smbert_v0000/smbert.onnx',
      Checksum(
          'c07653a3cc568876071de157f0fb8ccc1a302197ddad9ead069a86140e19d913'),
      [
[...]

PR #226, opened 2021-09-23
TY-2107 load genesis.js in dart

Robert-Steiner:
Ticket:
- [TY-2107]

Based on:  #222 

[TY-2107]: https://xainag.atlassian.net/browse/TY-2107

### DIFF
--- a/bindings/dart/example/lib/data_provider/web.dart
+++ b/bindings/dart/example/lib/data_provider/web.dart
@@ -1,5 +1,7 @@
-import 'dart:html' show window;
-import 'dart:typed_data' show Uint8List, ByteBuffer;
+import 'dart:async' show Completer;
+import 'dart:html' show document, window;
+import 'dart:js' show context;
+import 'dart:typed_data' show ByteBuffer, Uint8List;
 
 import 'package:xayn_ai_ffi_dart/package.dart'
     show AssetType, getAssets, SetupData, WebFeature;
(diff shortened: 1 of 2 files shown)


bindings/dart/example/lib/data_provider/web.dart
```diff
@@ -43,3 +47,24 @@ Future<Uint8List> _fetchAsset(String url, String checksum) async {
...
+/// Injects the WASM script into the HTML and exports its methods
+/// to the `window` object of the browser.
+Future<void> injectWasmScript(Uint8List scriptBytes) {
+  final completer = Completer<void>();
+  final src =
+      Url.createObjectUrl(Blob(<Uint8List>[scriptBytes], 'text/javascript'));
+
+  context['_notifyWasmBindingsLoaded'] = () => completer.complete();
+
+  document.head!.append(document.createElement('script')
```
janpetschexain, line 60: not sure if i understand this correctly, the script used to be in the html body before, does injecting it into the head has the same effect?
```suggestion
  document.body!.append(document.createElement('script')
```

Robert-Steiner: In our case it should have the same effect. It would matter [if the script execution takes awhile to run or if the script depends on the DOM](https://medium.com/geekculture/where-to-put-a-script-tag-into-head-or-body-end-b5b063058e0b).

janpetschexain (approved):

PR #266, opened 2021-10-28
TY 2079 [1] System takes documents as immutable 

acrrd:
Since now we want to have fallback strategies we cannot take ownership of the documents in the systems.
Systems are now taking a slice.

I didn't do it for the context system because it is the last one that gets the data and there is not fallback for it.

This is based on #265, only the last commit belongs to this PR.

### DIFF
--- a/xayn-ai/src/ltr/mod.rs
+++ b/xayn-ai/src/ltr/mod.rs
@@ -32,7 +32,7 @@ impl LtrSystem for DomainReranker {
     fn compute_ltr(
         &self,
         history: &[DocumentHistory],
-        documents: Vec<DocumentDataWithQAMBert>,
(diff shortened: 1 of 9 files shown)


xayn-ai/src/ltr/mod.rs
```diff
@@ -247,7 +247,7 @@ mod tests {
             coi,
         };
 
-        let res = ConstLtr.compute_ltr(&[], vec![doc1, doc2]);
+        let res = ConstLtr.compute_ltr(&[], &vec![doc1, doc2]);
```
Robert-Steiner, line 250:
```suggestion
        let res = ConstLtr.compute_ltr(&[], &[doc1, doc2]);
```

Robert-Steiner (approved):

PR #267, opened 2021-10-28
TY 2079 [2] Rerank with ok systems

acrrd:
Rerank using the scores from the systems that can run.

This is based on #266, the last commit belongs to this PR.

### DIFF
--- a/xayn-ai/src/reranker/mod.rs
+++ b/xayn-ai/src/reranker/mod.rs
@@ -823,7 +823,6 @@ mod tests {
         });
 
         let mut reranker = Reranker::new(cs).unwrap();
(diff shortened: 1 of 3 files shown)


xayn-ai/src/reranker/mod.rs
```diff
@@ -822,12 +822,35 @@ mod tests {
             coi
         });
 
-        test_system_failure(cs, true, mode);
+        let mut reranker = Reranker::new(cs).unwrap();
+        // let documents = documents_from_ids(0..10);
```
Robert-Steiner, line 826: Can we remove that?
```suggestion
```

Robert-Steiner (approved):

PR #269, opened 2021-10-29
[Merged by Bors] - TY-2136 impl kpe models manually [1]

janpetschexain:
**References**

- [TY-2136]

**Summary**

i'd suggest to look at the [commit of the raw impl](https://github.com/xaynetwork/xayn_ai/blob/b692b68714776f3ddcaee70cd48f8e64875c28b8/layer/src/conv.rs) to check correctness wrt the original C++ version and then at the last commit to check if this was rustified correctly:
- add a 1D convolutional layer and implement the paths needed from the [C++ torch source](https://pytorch.org/docs/stable/generated/torch.nn.functional.conv1d.html) for the chosen configuration in the [KPE prototype](https://github.com/xaynetwork/xayn_ai_research/blob/main/experiments/preference_modeling/keyphrase_extraction/bert2joint/model.py#L193-L195)
- add tests, the original only had one test and i created more from running the original

**Todo**

- ~refactor the looped assignments as iterator collections whereever possible~

[TY-2136]: https://xainag.atlassian.net/browse/TY-2136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/layer/src/conv.rs
+++ b/layer/src/conv.rs
@@ -261,7 +261,7 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "not implemented: ATen/native/")]
+    #[should_panic(expected = "not implemented: ATen/native/cpu/Unfold2d")]
     fn test_conv1d_padding_with_nonsingleton_kernel() {
         let weights = Array1::range(0., 18., 1.).into_shape((2, 3, 3)).unwrap();
         let input = Array1::range(0., 30., 1.).into_shape((2, 3, 5)).unwrap();
@@ -283,7 +283,7 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "not implemented: ATen/native/")]
+    #[should_panic(expected = "not implemented: ATen/native/cpu/Unfold2d")]
     fn test_conv1d_padding_with_singleton_kernel() {
         let weights = Array1::range(0., 6., 1.).into_shape((2, 3, 1)).unwrap();
         let input = Array1::range(0., 30., 1.).into_shape((2, 3, 5)).unwrap();
@@ -305,7 +305,7 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "not implemented: ATen/native/")]
+    #[should_panic(expected = "not implemented: ATen/native/NaiveDilatedConvolution")]
     fn test_conv1d_dilation() {
         let weights = Array1::range(0., 18., 1.).into_shape((2, 3, 3)).unwrap();
         let input = Array1::range(0., 30., 1.).into_shape((2, 3, 5)).unwrap();
@@ -318,7 +318,7 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "not implemented: ATen/native/")]
+    #[should_panic(expected = "not implemented: ATen/native/Convolution")]
     fn test_conv1d_groups() {
         let weights = Array1::range(0., 24., 1.).into_shape((2, 4, 3)).unwrap();
         let input = Array1::range(0., 80., 1.).into_shape((2, 8, 5)).unwrap();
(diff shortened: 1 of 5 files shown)


layer/src/conv.rs
```diff
@@ -0,0 +1,351 @@
...
+
+use displaydoc::Display;
+use ndarray::{s, Array1, Array2, Array3, ArrayBase, Data, ErrorKind, Ix2, Ix3, ShapeError};
+use thiserror::Error;
+
+/// A 1-dimensional convolutional layer.
+pub struct Conv1D {
+    // params
+    weights: Array2<f32>,
+    bias: Option<Array3<f32>>,
```
rustonaut, line 11:
```suggestion
    weights: Array3<f32>,
    bias: Option<Array1<f32>>,
```

I still need to double check it but bias being 3-D and with that higher dimensional then weights seems strange, tbh.

rustonaut (changes requested):
Besides the one dimension mess up one problem is that the doc-links are not perma links, i.e.  they link to master and when master changes might point to pretty arbitrary places

janpetschexain:
> Besides the one dimension mess up one problem is that the doc-links are not perma links, i.e. they link to master and when master changes might point to pretty arbitrary places

i want to remove/replace the doclinks once the review is done, they are just here to help reviewing the code.

layer/src/conv.rs
janpetschexain: the weights are already reshaped into the dimension needed during the run, such that we don't have to reshape them every time.

the bias is just extended with singleton dimensions from `(bias_size,)` to `(1, bias_size, 1)` for the same reason.

rustonaut: Oh, you are right I even preciously checked that the transformation was correct :see_no_evil:

layer/src/conv.rs
```diff
@@ -75,21 +81,8 @@ impl Conv1D {
...
         let dilated_kernel_size = dilation * (kernel_size - 1) + 1;
 
-        if groups > channel_out_size {
-            return Err(ConvError::Config(format!(
-                "given groups={}, expected weight to be at least {} at dimension 0, but got weight of size {:?} instead",
-                groups,
-                groups,
-                weights_shape,
-            )));
-        }
```
rustonaut: is it ok to remove this?

rustonaut (commented):
-

layer/src/conv.rs
janpetschexain: yes, it's already implied: both `groups` and `channel_out_size` are guaranteed to be non-zero, hence any `channel_out_size < groups` will cause `channel_out_size % groups == channel_out_size != 0` wich is already checked below.

rustonaut (approved):

janpetschexain:
@rustonaut why are you merging my pr?

rustonaut:
Isn't it the new PR flow? Given that bors by default doesn't allow the PR opener to merge it?

rustonaut:
I mane `borse r+` ==  “reviewed: looks good.”

janpetschexain:
> Isn't it the new PR flow? Given that bors by default doesn't allow the PR opener to merge it?

why should it be? the pr opener can use bors to merge the pr as soon as the pr has one approval, see the list of merged prs of the repo. anything else would be pretty weird.

> I mane borse r+ == “reviewed: looks good.”

actually `bors r+` is just shorthand for `bors merge`, so it's not a "reviewed: looks good" but a "merge pr".

rustonaut:
> actually bors r+ is just shorthand for bors merge, so it's not a "reviewed: looks good" but a "merge pr".

It is a quote from our own documentation on Confluence.

janpetschexain:
> > actually bors r+ is just shorthand for bors merge, so it's not a "reviewed: looks good" but a "merge pr".
> 
> It is a quote from our own documentation on Confluence.

then we should fix that documentation.

PR #271, opened 2021-11-01
[Merged by Bors] - DEVOPS-590 Bors integration for the CI

jorge-cr-13:
This file will integrate bors with the repository, bors will run with the CI file. All the jobs on the CI have to run successfully in order for bors to squash and merge a PR.

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -1,4 +1,4 @@
 status = [ "ci-ok"]
 timeout_sec = 18000 # five hours
-required_approvals = 1
+#required_approvals = 1
 use_squash_merge = true
--- a/bors.toml
+++ b/bors.toml
@@ -1,4 +1,4 @@
-status = [ "cargo-registry-cache", "cargo-format", "cargo-sort", "cargo-check", "cargo-clippy","cargo-test", "cargo-tarpaulin","test-android-libs","build-ios-libs", "test-wasm-libs","build-linux-lib","flutter-format","flutter-checks","flutter-build-example"]
+status = [ "ci-ok"]
 timeout_sec = 36000 # ten hours
 required_approvals = 1
 use_squash_merge = true
--- a/bors.toml
+++ b/bors.toml
@@ -1,0 +1,4 @@
+status = [ "cargo-registry-cache", "cargo-format", "cargo-sort", "cargo-check", "cargo-clippy","cargo-test", "cargo-tarpaulin","test-android-libs","build-ios-libs", "test-wasm-libs","build-linux-lib","flutter-format","flutter-checks","flutter-build-example"]
+timeout_sec = 36000 # ten hours
+required_approvals = 1
+use_squash_merge = true
--- a/bors.toml
+++ b/bors.toml
@@ -1,4 +1,4 @@
 status = [ "ci-ok"]
 timeout_sec = 18000 # five hours
-#required_approvals = 1
+required_approvals = 1
 use_squash_merge = true
--- a/bors.toml
+++ b/bors.toml
@@ -1,4 +1,4 @@
 status = [ "ci-ok"]
-timeout_sec = 36000 # ten hours
+timeout_sec = 18000 # five hours
 required_approvals = 1
 use_squash_merge = true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -676,8 +676,8 @@ jobs:
   # this is an helper that needs all the real leafs of the workflow.
   # It makes easier notify_staging_failure because we only need to check
   # for this job
-  one-leaf:
-    name: ignore-me
+  ci-ok:
+    name: ci-ok
     needs:
       - cargo-sort
       - flutter-format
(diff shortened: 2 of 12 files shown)


bors.toml
```diff
@@ -0,0 +1,4 @@
+status = [ "cargo-registry-cache", "cargo-format", "cargo-sort", "cargo-check", "cargo-clippy","cargo-test", "cargo-tarpaulin","test-android-libs","build-ios-libs", "test-wasm-libs","build-linux-lib","flutter-format","flutter-checks","flutter-build-example"]
```
acrrd, line 1: Do we have to least them all or we can a leaf job in the CI graph called `bors_ok` or somthing and only list that here? Can you please take a look what other projects are doing?

bors.toml
```diff
@@ -0,0 +1,4 @@
...
+timeout_sec = 36000 # ten hours
```
acrrd, line 2: This is so high because the timeouts in the CI are working and is to avoid to update this evry time we need to update a timeout in the ci?

acrrd (commented):
Will bors already work for this pr?

bors.toml
jorge-cr-13: Yes, we can make a big job checking the status of all of the others and then if it is ok to merge it. I think it can look more orderly that way.

bors.toml
jorge-cr-13: Exactly, if any of the previous status fail then it will fail. So it isn't so critical. It is just if in case it gets stuck trying to merge, then it will fail. But I can do it less? But how less? 5 hours?

acrrd: ok, let's do 5

Robert-Steiner:
There is also a `required_approvals` option. Not sure if without it we can trigger bros without an approval.

acrrd:
Let's try.

@bors try

bors.toml
acrrd: We actually already have one: `one-leaf`, it is used to simplify a bit `notify-failure`.
We can rename it `ci-ok` and use it for both `notify-failure` and here for bors.

acrrd:
Also try needs and approval?

jorge-cr-13:
bors try

jorge-cr-13:
> Also try needs and approval?

No, but it is without the @.

abargiela (approved):

jorge-cr-13:
> > ## try
> > Timed out.
> 
> @jorge-cr-13 can you take a look?

I would say that it is because I made changes to the configuration while it was running, but not sure because the action is finished correctly. So I'm trying again to see if it still times out.

jorge-cr-13:
borse try

jorge-cr-13:
bors try

jorge-cr-13:
> ## try
> Build succeeded:
> 
> * [ci-ok](https://github.com/xaynetwork/xayn_ai/runs/4099090383?check_suite_focus=true)

@acrrd I believe the problem with the first try, was that I changed the configuration of it while bors was still running. Also, I checked and you can't link the GH action on the commit message.

acrrd:
cool, thanks.

Let's give it go

bors r+

jorge-cr-13:
> This PR was included in a batch that successfully built, but then failed to merge into master. It will not be retried.
> 
> Additional information:
> 
> ```json
> {"message":"You're not authorized to push to this branch. Visit https://docs.github.com/articles/about-protected-branches/ for more information.","documentation_url":"https://docs.github.com/articles/about-protected-branches"}
> ```

@acrrd we have to add bors as one of the accounts to be able to merge into master, what do you think?

acrrd:
> @acrrd we have to add bors as one of the accounts to be able to merge into master, what do you think?

Yep, sure. Let's try to give it as less permission as possible, but if we want to use it we have no choice :)

acrrd:
> This PR was included in a batch that successfully built, but then failed to merge into master. It will not be retried.
> 
> Additional information:
> 
> ```json
> {"message":"At least 1 approving review is required by reviewers with write access.","documentation_url":"https://docs.github.com/articles/about-protected-branches"}
> ```

We don't remove approval when new code is pushed, but maybe @abargiela don't have write permission?

abargiela (approved):

Robert-Steiner:
@jorge-cr-13 do you know why bors is always running the ci for two branches when we do `+r` ?
Screen Shot 2021-11-04 at 16 57 23

acrrd:
> We could change the CI to ignore this branch and that way it will only run once, I believe running the full CI twice is too much. Wdyt?

Sounds good

jorge-cr-13:
> > We could change the CI to ignore this branch and that way it will only run once, I believe running the full CI twice is too much. Wdyt?
> 
> Sounds good

Also, I'm going to remove the required_approval from the bors file. Anyways is already set up under branch protection, so not even bors will be able to merge if there is no approval. I read that bors and branch protection don't work so good together, so I'm going to try with just one setup.

acrrd:
> Also, I'm going to remove the required_approval from the bors file. Anyways is already set up under branch protection, so not even bors will be able to merge if there is no approval. I read that bors and branch protection don't work so good together, so I'm going to try with just one setup.

Could be then that github is forbidding the merge because the branch from which bors is merging never received an approval?

acrrd:
try r+ without require approval in bors
bors r+

PR #285, opened 2021-11-08
[Merged by Bors] - DEVOPS-602 Changes labeler config file

jorge-cr-13:
Standarized the labeler configuration using commit versions and removing the if statements.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -10,37 +10,38 @@ on:
 jobs:
   draft_PR:
     if: (github.event.pull_request.draft == true)
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     name: Add WIP label
     steps:
       - name: Add WIP Label
-        uses: buildsville/add-remove-label@v1
+        uses: buildsville/add-remove-label@6008d7bd99d3baeb7c04033584e68f8ec80b198b # v1
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/labeler.yml
```diff
@@ -10,37 +10,38 @@ on:
...
           token: ${{secrets.LABEL_MODIFIER_TOKEN_GITHUB}}
           label: "WIP ⏳"
           type: add
+
       - name: Remove Ready for Review Label 
-        uses: buildsville/add-remove-label@v1
-        if: contains(github.event.pull_request.labels.*.name, 'Ready for review ✅')
+        uses: buildsville/add-remove-label@6008d7bd99d3baeb7c04033584e68f8ec80b198b # v1
         with:
           token: ${{secrets.LABEL_MODIFIER_TOKEN_GITHUB}}
```
Robert-Steiner, line 26: Just for my understanding 
For what reason are we using here an extra token instead of the `GITHUB_TOKEN`?

Robert-Steiner (approved):

.github/workflows/labeler.yml
jorge-cr-13: To be honest, when I created the CI I didn't think about this option. I will test with the `GITHUB_TOKEN` and if it works fine, I don't know if it has the same permission, then I would change it everywhere to make it more standard.

PR #286, opened 2021-11-09
[Merged by Bors] - DEVOPS-600 Creating a new CI for bors

jorge-cr-13:
Splitting the main CI into 2 different sections, one to run normally when PR are created and the other one for bors to merge.

### DIFF
--- a/.github/workflows/bors_ci.yml
+++ b/.github/workflows/bors_ci.yml
@@ -1,16 +1,12 @@
 # You can find more information about the design decisions here:
 # https://github.com/xaynetwork/xayn_ai/pull/21
 
-name: Rust-CI
+name: Bors-Rust-CI
 
 on:
   push:
-    branches-ignore:
-      # we push on release only from staging and release represent
-      # a snapshot of staging in a given point in time
-      - 'master'
-      - 'release'
-      - '_bors_staging-squash-merge.tmp'
+    branches:
+    - 'master'
 
 concurrency:
   group: ${{ github.ref }}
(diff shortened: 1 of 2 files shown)


.github/workflows/bors_ci.yml
```diff
@@ -0,0 +1,711 @@
+# You can find more information about the design decisions here:
+# https://github.com/xaynetwork/xayn_ai/pull/21
+
+name: Bors-Rust-CI
```
acrrd, line 4:
```suggestion
name: Merge master CI
```
seems a bit better. And we can rename the other ci something like PR CI or Dev CI.

.github/workflows/ci.yml
```diff
@@ -288,153 +289,6 @@ jobs:
...
-      - name: Download data
-        run: sh download_data.sh
-
-      - name: Test WASM library
-        working-directory: ./xayn-ai-ffi-wasm
-        env:
-          RUSTFLAGS: ${{ env.OPT_TESTS_RUSTFLAGS }}
-        run: wasm-pack test --firefox --chrome --headless
-
   build-linux-lib:
```
acrrd, line 292: Can't we remove also this? This job builds the libraris to be used in flutter-build-example for android.

acrrd:
bors try

.github/workflows/ci.yml
Robert-Steiner: Yes, but we still need to run the build scripts in order to be able to generate the ffi bindings otherwise the analyzer will fail. 
we could get the headers from `cargo check` and remove that job

acrrd: ah you are right, we need to keep it then because we are running the tests also there.

acrrd (approved):

jorge-cr-13:
bors try+

PR #291, opened 2021-11-11
Discussion about new linters

acrrd:
New linters are defined here `xayn-ai/src/lib.rs` and copied below:
```rust
// all warning are errors, we can enforce this here instead of -D warning on the command line
#![deny(warnings)]
#![deny(
    // pub items must have documentation
    missing_docs,
    // warn in case like &T.clone() where T: !Send, in this case the reference will be cloned
    // but is not the intended meaning
    noop_method_call,
    // some warning are error in the compiler with 1.56 but there are still warn lints that we
    // need to fix
    rust_2018_idioms,
    // warn about code not compatible with edition 2021
    // rust_2021_compatibility, we can only use it when we switch to 1.55 and remove it after 1.56
    // pub items that are not reachable from outside and thus should be pub(crate)
    // sealed trait can set this to allow
    unreachable_pub,
    // we can remove this for crates that need to use unsafe (ffi)
    unsafe_code,
    unsafe_op_in_unsafe_fn,
    // warn when we import a name but we use it qualified: use foo::bar; fn f() { foo::bar }
    unused_qualifications,
)]
// enable a lot of linters
// most warn that we have at the moment:
// use _ in numbers. 10000 => 10_000
// variable name to similar: doc & dom, coi1 & cois
// struct/enum name starts/end with the name of the module documents.rs => DocumentId
// we can decide to allows some of these if they are too annoying
#![deny(clippy::pedantic)]
// if future is not send it cannot be sent to another thread reducing performance
#![deny(clippy::future_not_send)]
```

here you can discover more about clippy lints:
https://rust-lang.github.io/rust-clippy/master/
and here about the lints from the compiler:
https://doc.rust-lang.org/rustc/lints/groups.html

We can start using this in new projects and switch to them here more slowly.

other commits:
one commit fix some of them using `cargo fix` (mainly rust_2018_idiom)
another one fix only a couple a lint from clippy,
In the last I fixed some of them manually do give more visibility to the errors the other one are producing.

This PR is meant to be merged.


janpetschexain:
i like the proposed lints. in some occasions where we want to enforce them globally without the possibility of (accidentally) disabling them, i'd suggest to even `forbid` them instead of just `deny` them. for example for the warnings, we should `#![forbid(warnings)]` instead of `#![deny(warnings)]`.

acrrd:
> for example for the warnings, we should `#![forbid(warnings)]` instead of `#![deny(warnings)]`.

for me is ok to use forbid for warnings but is always possible to reduce the level with `--cap-lints`.

acrrd:
My concern is if denying warning will hurt local development. While writing code locally can be ok to have warning while trying stuff

Robert-Steiner:
> My concern is if denying warning will hurt local development. While writing code locally can be ok to have warning while trying stuff

So does it mean if the code produces `warnings` the compilation will fail?

acrrd:
yes, every warning will be an error instead

Robert-Steiner:
> yes, every warning will be an error instead

ahh ok :/ yeah that could be annoying during development/testing

rustonaut:
> #![deny(warnings)]

Is considered a anti-patterns (I think in general, but at least by me) as it brakes semver guarantees, the compiler and clippy are always free to introduce new warnings. Code should not "arbitrarily" fail with a newer rust compiler just because the compiler introduced a new warning/or has a false positive warning. 

I personally am pretty strongly against `deny(warnings)` at most I would use it in borse by injecting it through a cmd-line argument. But even in such a setup I it can cause headaches when you e.g. need to do some quick fixes while some dependencies/tools changes.

> I'd suggest to even forbid them

I'm not a bit fan of `forbid`ing anything which you might want to temporary allow during dev, or even beyond for some edge-cases.  At least if you have reviews it's not likely to "accidentally" disable some lints. The only situation where I ran into "accidental disabled lints" was in combination with `deny(warnings)`, which like I already mentioned I consider an anti pattern.

So I would not use `forbid` for any of:  `missing_docs`, `warnings`, `unreachable_pub` and `unsafe_code`.

Through tbh. in personal projects I likely wouldn't touch it at all as it's not worth the hassle.

> // if future is not send it cannot be sent to another thread reducing performance
> #![deny(clippy::future_not_send)]

I have nothing against this lint but the description in the comment is a bit misleading.

A future not being send is less about performance and more about usability, as `Send` futures can be used with more async functions and  spawn functions then `!Send` futures and the `spawn` method used is in the end which determines certain performance aspects. But then under some circumstances a `!Send` future could be noticeable more performant then a `Send` future. Either way in my experience in most cases of futures not being send it is a accidental/undesired, so `deny` or `warn`-ing it is a good idea (just not `forbid`ding it).

PR #300, opened 2021-11-15
DEVOPS-603 Cargo format ci

jorge-cr-13:
Creating new composite actions with the different jobs of the CI.


acrrd:
This is a test with cargo-format but we can extend this to all jobs in common between `ci.yml` and `bors_ci.yml`. The alternative would be to use the reusable workflow and call `ci.yml` from `bors_ci.yml` but is a feature that is still in beta. We can also wait till that is a bit more stable if we want.

acrrd:
@jorge-cr-13  can you please try to use this to see if we can have subgroup in the actions? https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#grouping-log-lines

Robert-Steiner:
> This is a test with cargo-format but we can extend this to all jobs in common between `ci.yml` and `bors_ci.yml`. The alternative would be to use the reusable workflow and call `ci.yml` from `bors_ci.yml` but is a feature that is still in beta. We can also wait till that is a bit more stable if we want.

If ci.yml is part of bors.yml, we could try reusable workflows.

A nice feature of reusable workflows is [`secrets`](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onworkflow_callsecrets). I haven't tried it yet but  my guess it that it takes care that the secrets are masked in the logs. That is not the case for composite actions. If an input value (token, key, etc) is not defined as a repo secret, it will be printed unmasked in the logs e.g.

Screen Shot 2021-11-15 at 17 33 27
You would see the keys here if they weren't in the repo secrets. 

Later we can still outsource parts into compound actions (as long as they work in reusable workflows, maybe we should try that first).

jorge-cr-13:
> @jorge-cr-13 can you please try to use this to see if we can have subgroup in the actions? https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#grouping-log-lines

The problem with this approach is that you have to add the grouping inside a command and most of the steps here are external actions. I tried encapsulating the step I wanted to group between steps with the grouping commands, but the grouping statements have to be in the same step.

PR #301, opened 2021-11-16
[Merged by Bors] - TY-2218 fix wasm RuntimeError handling

Robert-Steiner:
Ticket:

[TY-2218]

Summary:

I noticed this bug when I was trying to catch a `WebAssembly.RuntimeError` in `_XaynAi`. 
The code failed in the `on XaynAiError catch` block with following message:
`Web worker error while handling the method call Method.create: Undefined enum variant.`

There were two problems: 

1. The `RuntimeError` was interpreted as a `XaynAiError`. When Dart translate the code into JS it generates some like this for the `on XaynAiError catch` line

```js
if (type$.XaynAiError._is(t1))
```  

The same JS code is generated if you write:  

```dart
catch (error) {
	if (error is XaynAiError)
```

And here lies the problem, since `XaynAiError` is a `@JS()` type, `is` always returns `true`

From the Readme of the [js package](https://pub.dev/packages/js)

> `is` checks and `as` casts between JS interop types will always succeed
>
> For any two `@JS()` types, with or without `@anonymous`, a check of whether an object of one type `is` another type > will always return true, regardless of whether those two types are in the same prototype chain. Similarly, an explicit cast  using `as` will also succeed. 

2. Our Dart type `RuntimeError` did not refer to the `WebAssembly.RuntimeError`, instead Dart creates a new type for that (`V.RuntimeError0`). I found the solution to this problem in the [`wasm.dart`](https://github.com/lexaknyazev/wasm.dart/blob/a6c93afea4732c140f1f61f144795961c42c8613/wasm_interop/lib/wasm_interop.dart#L718) repo. 

The generated JS error handling code before the change:

```js
catch (exception) {
	t1 = H.unwrapException(exception);
	if (type$.XaynAiError._is(t1)) {
    	        error = t1;
    	        throw H.wrapException(V.XaynAiErrorToException_toException(error));
	} else if (t1 instanceof V.RuntimeError0) {
		error0 = t1;
		throw H.wrapException(V.RuntimeErrorToException_toException(error0));
	} else
		throw exception;
	}
```

and after:

```js
catch (exception) {
	error = H.unwrapException(exception);
	if (error instanceof self.WebAssembly.RuntimeError)
		throw H.wrapException(V.RuntimeErrorToException_toException(V.ObjectToRuntimeError_toRuntimeError(error)));
	else {
		t1 = error;
		if ("code" in t1 && "message" in t1)
			throw H.wrapException(V.XaynAiErrorToException_toException(V.ObjectToXaynAiError_toXaynAiError(error)));
		else
			throw exception;
		}
	}
```

The error `Undefined enum variant` was thrown in the `toCode` function because the value of `code` was `undefined` (it does not exist in `RuntimeError`).

Tested `flutter run`, `flutter build` and `flutter build --release` on Chrome, Firefox and Safari.

[TY-2218]: https://xainag.atlassian.net/browse/TY-2218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bindings/dart/lib/src/web/ffi/ai.dart
+++ b/bindings/dart/lib/src/web/ffi/ai.dart
@@ -4,6 +4,7 @@ library ai;
 import 'dart:typed_data' show Uint8List;
 
 import 'package:js/js.dart' show JS;
+import 'package:js/js_util.dart' show instanceof;
 
 import 'package:xayn_ai_ffi_dart/src/common/data/document.dart' show Document;
 import 'package:xayn_ai_ffi_dart/src/common/data/history.dart' show History;
@@ -22,10 +23,12 @@ import 'package:xayn_ai_ffi_dart/src/web/reranker/analytics.dart'
     show JsAnalytics, ToAnalytics;
 import 'package:xayn_ai_ffi_dart/src/web/result/error.dart'
     show
-        RuntimeError,
+        ObjectToRuntimeError,
+        ObjectToXaynAiError,
         RuntimeErrorToException,
         XaynAiError,
-        XaynAiErrorToException;
+        XaynAiErrorToException,
+        runtimeError;
 import 'package:xayn_ai_ffi_dart/src/web/result/fault.dart'
     show JsFault, ToStrings;
 import 'package:xayn_ai_ffi_dart/src/web/result/outcomes.dart'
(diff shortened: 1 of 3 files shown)


janpetschexain (approved):
nice fix 👍

acrrd (approved):

Robert-Steiner:
bors merge

PR #306, opened 2021-11-18
[Merged by Bors] - TY-2136 impl kpe models manually [6]

janpetschexain:
**References**

- #302

**Summary**

some follow up cleanup for `rubert` noticed during #302:
- update docs: tract's internals changed & additional shape info
- constify model shapes and add debug assertions: the embedding size is (de facto) fixed per model kind and this gives us additional shape safety during model creation
- restrict token size range: this is inherently given by the onnx models
- use ndarray's shape error instead of a custom one

the additional exposure of the `BertModel` trait doesn't allow it to be hidden anymore (at least without a lot of cumbersome wrapping), it is still unimplementable from outside the crate, but i linked it in the docs. this also allows us to remove the `Model::new()` wrapper. i'd say that the additional shape safety is worth the trait doc exposure.

### DIFF
--- a/rubert/src/builder.rs
+++ b/rubert/src/builder.rs
@@ -88,13 +88,16 @@ impl<V, M, K, P> Builder<V, M, K, P> {
     /// Defaults to `128`.
     ///
     /// # Errors
-    /// Fails if `size` is less than two.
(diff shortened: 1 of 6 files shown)


rustonaut (approved):
lgtm

janpetschexain:
bors merge

PR #308, opened 2021-11-19
[Merged by Bors] - FIX: tokenizer offsets

janpetschexain:
**Summary**

- fix the token offsets computed during the tokenizer's encoding: some special tokens had default offsets, but their offsets must be aligned with the offsets of their surrounding tokens
- expose `Offsets`: this is already part of the public api, but it was not nameable


finiteprods (approved):
lgtm

janpetschexain:
bors merge

PR #327, opened 2021-11-29
[Merged by Bors] - Remove cargo-check. Allow jobs to run earlier

acrrd:
This PR aims to speed up the ci allowing some job to run earlier.

* Remove cargo-check. Compilation errors will be present in cargo-clippy and cargo-test
* Split cargo-test in two:
  * cargo-test that runs only `cargo test --all-features` 
  * cargo-more-test that runs other tests with --all-target and also --doc.
the jobs `cargo-tarpaulin` and `build-linux-lib` now depends only on `cargo-test`.

### DIFF
--- a/.github/workflows/ci_reusable_wf.yml
+++ b/.github/workflows/ci_reusable_wf.yml
@@ -16,7 +16,6 @@ env:
 
 jobs:
   cargo-format:
-    name: cargo-format
     runs-on: ubuntu-20.04
     timeout-minutes: 10
     steps:
@@ -39,7 +38,6 @@ jobs:
         run: cargo fmt --all -- --check
 
   cargo-sort:
-    name: cargo-sort
     runs-on: ubuntu-20.04
     timeout-minutes: 10
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_reusable_wf.yml
```diff
@@ -164,21 +164,31 @@ jobs:
...
+          RUSTFLAGS: "-D warnings"
         run: |
-          cargo clippy --all-targets -- --deny warnings
-          cargo clippy --all-targets --all-features -- --deny warnings
-          
-  cargo-test:
-    name: cargo-test
-    needs: [cargo-registry-cache, cargo-check]
+          # compilations options must be kept in sync with cargo-tarpaulin and build-linux-lib
+          cargo test --all-features
```
janpetschexain, line 187: since run consists of only one line now, we could remove the pipe
```suggestion
        # compilations options must be kept in sync with cargo-tarpaulin and build-linux-lib
        run: cargo test --all-features
```

.github/workflows/ci_reusable_wf.yml
```diff
@@ -164,21 +164,31 @@ jobs:
...
+        env:
+          RUSTFLAGS: "-D warnings"
         run: |
-          cargo clippy --all-targets -- --deny warnings
-          cargo clippy --all-targets --all-features -- --deny warnings
-          
-  cargo-test:
-    name: cargo-test
-    needs: [cargo-registry-cache, cargo-check]
+          # compilations options must be kept in sync with cargo-tarpaulin and build-linux-lib
```
janpetschexain, line 186: should we put the mentioned options in an env var and use that instead of having the comment?

.github/workflows/ci_reusable_wf.yml
```diff
@@ -321,7 +322,7 @@ jobs:
         env:
           RUSTFLAGS: "-D warnings"
         # compilations options must be kept in sync with cargo-test
-        run: cargo build --all-features
+        run: cargo build ${{ env.COMMON_CARGO_FLAG_TESTS }}
```
janpetschexain, line 325: typo
```suggestion
        run: cargo build ${{ env.COMMON_CARGO_TEST_FLAGS }}
```

janpetschexain (changes requested):

.github/workflows/ci_reusable_wf.yml
acrrd: so github does not complain if I refer to something that is not in `env`?

janpetschexain: i guess it returns an empty string if the env can't be found.

janpetschexain (approved):

acrrd:
cargo merge

acrrd:
> cargo merge

ops 😂

bors merge

PR #328, opened 2021-11-29
[Merged by Bors] - DEVOPS-599 Using the latest tag for the reusable WF

jorge-cr-13:
This tries to avoid having to change the reference from the reusable workflow everytime a change is implemented to it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ permissions:
 jobs:
 
   dev-ci:
-    uses: xaynetwork/xayn_ai/.github/workflows/ci_reusable_wf.yml@49cefaead3a29817658e7e19d1ea028070aa1f3b
+    uses: xaynetwork/xayn_ai/.github/workflows/ci_reusable_wf.yml@latest
 
   # this is an helper that needs all the real leafs of the workflow.
   # It makes easier notify_staging_failure because we only need to check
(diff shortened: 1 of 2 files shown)


janpetschexain (approved):

.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ permissions:
 jobs:
 
   dev-ci:
-    uses: xaynetwork/xayn_ai/.github/workflows/ci_reusable_wf.yml@49cefaead3a29817658e7e19d1ea028070aa1f3b
+    uses: xaynetwork/xayn_ai/.github/workflows/ci_reusable_wf.yml@master
```
acrrd, line 22: I'm not sure we want this. This will break other branches if the ci changes on master

jorge-cr-13: Why would it break the other branches? This is only to reference where the CI has to look for the latest version of the reusable file.

acrrd: This will not break anything. But in the future we want some branch like `staging` and `release` to stay the same even if modify the ci on `master`. I talked with @janpetschexain about this, we can merge this and optimize for the most common case. During the release procedure we need to change this tag to a specific commit. There is also the case when ci and code must change in the same PR for some reason, in this case we will need two PR but it should be rare that we cannot split it in two different pr.

jorge-cr-13:
bors merge

PR #330, opened 2021-12-01
[Merged by Bors] - TY-2205 add coi weighting fields [0]

janpetschexain:
**References**

- [TY-2205]

**Summary**

- add view statistics fields to `PositiveCoi` & update the version migration schemes
- update the `CoiPoint::new()` constructor with the required fields & move the constructor definitions to avoid code duplication
- add `CoiPoint::update_view()` to use the new fields in existing cois

**Note**

- in every call `viewed: None` is currently used, this will be changed in #331
- if you use rust-analyzer, then you should add the `derivative` crate to the `Cargo: Unset Test` setting option, otherwise you will get a false positive for `unresolved-import`
- on `wasm32` targets we don't have access to the current system time neither from `std::time` nor from other crates like `time` or `chrono`, hence we have to use the `wasm32` target specific time from `js-sys`

[TY-2205]: https://xainag.atlassian.net/browse/TY-2205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/xayn-ai/src/coi/point.rs
+++ b/xayn-ai/src/coi/point.rs
@@ -80,6 +80,9 @@ pub(crate) trait CoiPoint {
     fn set_id(&mut self, id: CoiId);
     fn point(&self) -> &Embedding;
     fn set_point(&mut self, embedding: Embedding);
+    fn update_view(&mut self, viewed: Option<Duration>) {
+        #![allow(unused_variables)]
+    }
 }
 
 macro_rules! impl_coi_point {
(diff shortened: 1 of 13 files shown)


acrrd (dismissed):

xayn-ai/src/coi/point.rs
```diff
@@ -47,65 +55,64 @@ impl From<PositiveCoi_v0_1_0> for PositiveCoi {
...
 pub(crate) trait CoiPoint {
-    fn new(id: CoiId, embedding: Embedding) -> Self;
+    fn new(id: CoiId, point: Embedding, viewed: Option<Duration>) -> Self;
     fn id(&self) -> CoiId;
     fn set_id(&mut self, id: CoiId);
     fn point(&self) -> &Embedding;
     fn set_point(&mut self, embedding: Embedding);
+    fn update_view(&mut self, viewed: Option<Duration>) {
+        #![allow(unused_variables)]
+    }
```
acrrd, line 85: Why not this?
```suggestion
    fn update_view(&mut self, _viewed: Option<Duration>) {}
```

acrrd (approved):

xayn-ai/src/coi/point.rs
janpetschexain: because then rust-analyzer shows `_viewed` as the field name hint also in the case where you provide a custom implementation, which might be confusing as it looks like this field would always be unused, but actually it's only unused in the default implementation.

janpetschexain:
bors merge

PR #334, opened 2021-12-02
[Merged by Bors] - FIX: flutter example benchmark visibility

janpetschexain:
**Summary**

- wrap the benchmark stats in a "stateful enum": cleanup and less logic in the main file
- add a progressing counter for the benchmark: improves the visual feedback in the flutter example app next to the console output

**Note**

- we need a tiny delay in the `setPending` callback after `setState` to give the interface time to actually refresh, the delay shouldn't be needed as this is the whole pupose of `setState`, but i tried different sync and async approaches and none of them worked without an additional delay

### DIFF
--- a/bindings/dart/example/lib/logic.dart
+++ b/bindings/dart/example/lib/logic.dart
@@ -134,7 +134,9 @@ class Logic {
     }
   }
 
-  Future<Stats> benchmark() async {
+  Future<Stats> benchmark(
+    Future<void> Function(int) setStateBenchmarkStatsPending,
+  ) async {
     print('Starting Benchmark');
 
     const preBenchNum = 10;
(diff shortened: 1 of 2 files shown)


bindings/dart/example/lib/logic.dart
```diff
@@ -200,3 +204,42 @@ class Outcome {
     return resultList;
   }
 }
+
+enum _BenchmarkStatsKind {
```
acrrd, line 208: Here I'll prefer to use inheritance 
```dart
abstract class BecnhmarkStats {
  static BecnhmarkStats none() { return _BecnhmarkStatsNone(); }
  static BecnhmarkStats pending(int iter) { return _BecnhmarkStatsIter(iter); }
  static BecnhmarkStats ready(Stats<num> stats) { ... }
}
```
We avoid `null` and we don't add too much boilerplate. It could be done with `freezed` but I don't think it is worth it.

acrrd (approved):

janpetschexain:
bors merge

janpetschexain:
bors merge

PR #337, opened 2021-12-06
[Merged by Bors] - Extend assert_approx_eq macro for f64

janpetschexain:
**Summary**

extend the `assert_approx_eq!` macro for `f64`:
- change the associated type `LeafElement` of `ApproxEqIter` to a generic type: the compiler couldn't infer the correct input type from the associated type in case of multiple impls (even with additional type hints, because the associated type only governs the output type), it worked so far because there was only the single `f32` implementation
- cleanup: some documentation and additional macro hygiene

[TY-2207]: https://xainag.atlassian.net/browse/TY-2207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test-utils/src/approx_eq.rs
+++ b/test-utils/src/approx_eq.rs
@@ -59,39 +59,35 @@ macro_rules! assert_approx_eq {
        $crate::assert_approx_eq!($t, $left, $right, epsilon = $epsilon, ulps = 2)
     };
     ($t:ty, $left:expr, $right:expr, epsilon = $epsilon:expr, ulps = $ulps:expr $(,)?) => {{
-       let epsilon = $epsilon;
+        let epsilon = $epsilon;
         let ulps = $ulps;
         let left = $left;
         let right = $right;
         let mut left_iter =
-            $crate::ApproxEqIter::indexed_iter_logical_order(&left, Vec::new());
+            $crate::ApproxEqIter::<$t>::indexed_iter_logical_order(&left, Vec::new());
         let mut right_iter =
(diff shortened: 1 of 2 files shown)


acrrd:
Why can't we use f32 in #338 ?

janpetschexain:
> Why can't we use f32 in #338 ?

we could use `f32` if we are ok with potentially losing precision. this really depends on how the weights are going to be used later on.

acrrd:
> > Why can't we use f32 in #338 ?
> 
> we could use `f32` if we are ok with potentially losing precision. this really depends on how the weights are going to be used later on.

It is not an issue. The granularity of a duration for us will be in seconds so everything will start with `.0` and we are doing only few operations. The weights will be used to pick more often keyphrases from the most weighted coi and to weight the score of a document but we don't need to much precision in these cases.

janpetschexain:
> It is not an issue. The granularity of a duration for us will be in seconds so everything will start with `.0` and we are doing only few operations. The weights will be used to pick more often keyphrases from the most weighted coi and to weight the score of a document but we don't need to much precision in these cases.

ok, makes sense, i'll change #338 to `f64`s.

i think it's still worthwhile to have this pr to extend our test utilities, but i will decouple both prs since they are not dependent on each other anymore.

acrrd (approved):

janpetschexain:
bors merge

janpetschexain:
bors merge

PR #339, opened 2021-12-09
[Merged by Bors] - TY-2208 update key phrases [0]

janpetschexain:
**References**

- [TY-2208]

**Summary**

- add `cosine_similarity()` and `pairwise_cosine_similarity()` utilities which will be used in the keyphrase selection
- use `ArrayBase` instead of `&Deref<ArrayBase>` in the utility function signatures: `ArrayBase` doesn't implement neither `AsRef` nor `Deref` because they are intended to be `view()`ed for a reference; as long as the utilities are just single standing functions the `Deref` approach works but it is much more ergonomic to combine utility functions with the `ArrayBase` approach

[TY-2208]: https://xainag.atlassian.net/browse/TY-2208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/xayn-ai/src/embedding/utils.rs
+++ b/xayn-ai/src/embedding/utils.rs
@@ -59,23 +59,24 @@ where
 pub fn pairwise_cosine_similarity<'a, I, S>(iter: I) -> Array2<f32>
 where
     I: IntoIterator<Item = &'a ArrayBase<S, Ix1>>,
-    I::IntoIter: Clone,
+    I::IntoIter: Clone + ExactSizeIterator,
     S: Data<Elem = f32> + 'a,
 {
     let iter = iter.into_iter();
+    let size = iter.len();
+
     let norms = iter.clone().map(|a| l2_norm(a.view())).collect::<Vec<_>>();
-    let size = iter.clone().count();
     let mut similarities = Array2::ones((size, size));
-    iter.clone().enumerate().for_each(|(i, a)| {
+    for (i, a) in iter.clone().enumerate() {
         if norms[i] != 0. {
-            iter.clone().enumerate().skip(i + 1).for_each(|(j, b)| {
+            for (j, b) in iter.clone().enumerate().skip(i + 1) {
                 if norms[j] != 0. {
                     similarities[[i, j]] = a.dot(b) / norms[i] / norms[j];
                     similarities[[j, i]] = similarities[[i, j]];
                 }
-            });
+            }
         }
-    });
+    }
 
     similarities
 }
--- a/xayn-ai/src/embedding/utils.rs
+++ b/xayn-ai/src/embedding/utils.rs
@@ -56,23 +56,26 @@ where
 /// # Panics
 /// Panics if the vectors don't consist solely of real values or their shapes don't match.
 #[allow(dead_code)]
-pub fn pairwise_cosine_similarity<S>(a: &[ArrayBase<S, Ix1>]) -> Array2<f32>
+pub fn pairwise_cosine_similarity<'a, I, S>(iter: I) -> Array2<f32>
 where
-    S: Data<Elem = f32>,
+    I: IntoIterator<Item = &'a ArrayBase<S, Ix1>>,
+    I::IntoIter: Clone,
+    S: Data<Elem = f32> + 'a,
 {
-    let norms = a.iter().map(|a| l2_norm(a.view())).collect::<Vec<_>>();
-    let size = a.len();
+    let iter = iter.into_iter();
+    let norms = iter.clone().map(|a| l2_norm(a.view())).collect::<Vec<_>>();
+    let size = iter.clone().count();
     let mut similarities = Array2::ones((size, size));
-    for i in 0..size {
+    iter.clone().enumerate().for_each(|(i, a)| {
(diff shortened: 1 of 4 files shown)


xayn-ai/src/embedding/utils.rs
```diff
@@ -1,90 +1,166 @@
...
+/// # Panics
+/// Panics if the vectors don't consist solely of real values or their shapes don't match.
+#[allow(dead_code)]
+pub fn pairwise_cosine_similarity<S>(a: &[ArrayBase<S, Ix1>]) -> Array2<f32>
+where
+    S: Data<Elem = f32>,
+{
+    let norms = a.iter().map(|a| l2_norm(a.view())).collect::<Vec<_>>();
+    let size = a.len();
+    let mut similarities = Array2::ones((size, size));
```
rustonaut, line 68: Just to be sure, the  problems with QEMU  only applies to `::zeros()`?

xayn-ai/src/embedding/utils.rs
```diff
@@ -1,90 +1,166 @@
...
+    for i in 0..size {
+        if norms[i] != 0. {
+            for j in i + 1..size {
+                if norms[j] != 0. {
+                    similarities[[i, j]] = a[i].dot(&a[j]) / norms[i] / norms[j];
+                    similarities[[j, i]] = similarities[[i, j]];
+                }
+            }
+        }
```
rustonaut, line 77: theoretically the else branch is mathematically undefined i.e. `NaN` is it ok to just set it to `1` for us? (I guess probably, but we maybe should document it).

rustonaut (commented):
just two questions otherwise it looks good

xayn-ai/src/embedding/utils.rs
janpetschexain: correct, only the `zeros` will lead to the `memset` path, all other values will lead to the `memcpy` path.

xayn-ai/src/embedding/utils.rs
```diff
@@ -56,23 +56,26 @@ where
 /// # Panics
 /// Panics if the vectors don't consist solely of real values or their shapes don't match.
 #[allow(dead_code)]
-pub fn pairwise_cosine_similarity<S>(a: &[ArrayBase<S, Ix1>]) -> Array2<f32>
+pub fn pairwise_cosine_similarity<'a, I, S>(iter: I) -> Array2<f32>
 where
-    S: Data<Elem = f32>,
+    I: IntoIterator<Item = &'a ArrayBase<S, Ix1>>,
+    I::IntoIter: Clone,
+    S: Data<Elem = f32> + 'a,
```
rustonaut, line 63: Do we need the additional complexity and could we not use `ExactSizedIterator`?

Both cloning and iterator and counting it a things prone to accidentally introducing a bunch of overhead.

xayn-ai/src/embedding/utils.rs
janpetschexain: if a norm of a vector is zero, then all its elements have to be zero as well. since the zero vector is pointing in all direction at the same time, it's ok to define the similarity between a vector and the zero vector as 1.

in the end we will not encounter this in the real world, because the compared embeddings are almost never zero vectors.

rustonaut (commented):
I'm not sure about your last change being a good idea, especially using `for_each` instead of `for` seems unnecessary.

xayn-ai/src/embedding/utils.rs
janpetschexain: i encountered a situation in the key phrase impl where it's nicer to have it accept an iterator to avoid collecting too much. the `ExactSizedIterator` is a good idea, i'll change that. cloning an iterator is usually quite cheap, for example the iterator for slices is just to pointers big.

janpetschexain:
> I'm not sure about your last change being a good idea, especially using `for_each` instead of `for` seems unnecessary.

in the end they are doing the very same thing, but i changed `for_each` back to `for`.

xayn-ai/src/embedding/utils.rs
rustonaut: > because the compared embeddings are almost never zero vectors.

ok, so we keep it that way

> since the zero vector is pointing in all direction at the same time,

Or no direction (it's e.g. not uncommon I think to define the 0-vector as orthogonal with all vectors, in which case it would default to 0).  I'm pretty sure it's mathematically undefined (in the sense that there is no agreement on that and it could be either or even neither) and the similarity is normally also not defined for `0` vectors. We can define it the way it is in the code, but then we should document it.

janpetschexain: fair point. i updated the docs accordingly.

in our case i prefer to define the similarity as 1 to keep it aligned with the pink code (although they are doing some improper computations, for the sake of avoiding division by zero, which results in the similarity to a zero vector to be 10^6  instead of 1).

rustonaut (approved):
ok, thanks.

janpetschexain:
bors merge

PR #347, opened 2021-12-14
[Merged by Bors] - TY-2206 add key phrases to coi

janpetschexain:
**References**

- [TY-2206]

**Summary**

- add `KeyPhrase` wrapper to hold key phrases and their embeddings in a coi
- add key phrases to `PositiveCoi`: ~the uniqueness invariant can't be enforced for example by a `HashSet` because `KeyPhrase` can't implement `Eq + Hash` and also not by sort & dedup because `KeyPhrase` can't implement `PartialOrd`/`Ord`~ the uniqueness is enforced only over the words of the key phrases, this is under the assumption that smbert gives a unique embedding for each unique key phrase and if we ever change the smbert we would need migration stategies anyways
- update `CoiPoint` setters/getters for key phrases, split out `CoiPointKeyPhrases` and `CoiPointStats`
- modify the bounds on `pairwise_cosine_similarity()`: follow up on #339
- add placeholders for key phrase updates: to be implemented in a follow up pr

[TY-2206]: https://xainag.atlassian.net/browse/TY-2206?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/xayn-ai/src/coi/point.rs
+++ b/xayn-ai/src/coi/point.rs
@@ -3,14 +3,22 @@ use std::{
     time::{Duration, SystemTime},
 };
 
-#[cfg(test)]
 use derivative::Derivative;
+use displaydoc::Display;
 use serde::{Deserialize, Serialize};
+use thiserror::Error;
 
 use crate::{coi::CoiId, embedding::utils::Embedding, utils::system_time_now};
 
-#[derive(Clone, Copy, Deserialize, Serialize)]
-#[cfg_attr(test, derive(Debug))]
+#[derive(Debug, Display, Error)]
+pub(crate) enum CoiError {
+    /// The key phrase is invalid (ie. either empty words or non-finite point)
+    InvalidKeyPhrase,
+    /// The key phrases are not unique
+    DuplicateKeyPhrases,
+}
+
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
 pub(crate) struct CoiStats {
     pub(crate) view_count: usize,
     pub(crate) view_time: Duration,
@@ -53,23 +61,41 @@ impl Default for CoiStats {
     }
 }
 
-#[derive(Clone, Deserialize, PartialEq, Serialize)]
-#[cfg_attr(test, derive(Debug))]
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 pub(crate) struct KeyPhrase {
-    pub(super) words: String,
-    pub(super) point: Embedding,
+    words: String,
+    point: Embedding,
 }
 
 impl KeyPhrase {
-    pub(super) fn is_valid(&self) -> bool {
-        !self.words.is_empty() && self.point.iter().copied().all(f32::is_finite)
+    #[allow(dead_code)]
+    pub(crate) fn new(words: String, point: Embedding) -> Result<Self, CoiError> {
+        if !words.is_empty() && point.iter().copied().all(f32::is_finite) {
+            Ok(Self { words, point })
+        } else {
+            Err(CoiError::InvalidKeyPhrase)
+        }
     }
+}
 
-    pub(super) fn is_unique(key_phrases: &[KeyPhrase]) -> bool {
-        key_phrases
-            .iter()
-            .enumerate()
-            .all(|(idx, this)| key_phrases.iter().skip(idx + 1).all(|other| this != other))
+// invariant: must be unique
+// note: can't use neither HashMap nor sort & dedup because the underlying ArrayBase doesn't
+// implement any of Eq, Hash, PartialOrd and Ord
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
+pub(crate) struct KeyPhrases(Vec<KeyPhrase>);
+
+impl KeyPhrases {
+    #[allow(dead_code)]
+    pub(crate) fn new(key_phrases: Vec<KeyPhrase>) -> Result<Self, CoiError> {
+        for (i, this) in key_phrases.iter().enumerate() {
+            for other in key_phrases[i + 1..].iter() {
+                if this == other {
+                    return Err(CoiError::DuplicateKeyPhrases);
+                }
+            }
+        }
+
+        Ok(Self(key_phrases))
     }
 }
 
@@ -78,17 +104,17 @@ impl KeyPhrase {
 #[obake(version("0.1.0"))]
 #[obake(version("0.2.0"))]
 #[obake(version("0.3.0"))]
-#[derive(Clone, Deserialize, Serialize)]
-#[cfg_attr(test, derive(Debug, Derivative), derivative(PartialEq))]
+#[derive(Clone, Debug, Derivative, Deserialize, Serialize)]
+#[derivative(PartialEq)]
 pub(crate) struct PositiveCoi {
     #[obake(cfg(">=0.0"))]
     pub(super) id: CoiId,
     #[obake(cfg(">=0.0"))]
     pub(super) point: Embedding,
     #[obake(cfg(">=0.3"))]
-    pub(super) key_phrases: Vec<KeyPhrase>, // invariant: key phrases must be unique
+    pub(super) key_phrases: KeyPhrases,
     #[obake(cfg(">=0.3"))]
-    #[cfg_attr(test, derivative(PartialEq = "ignore"))]
+    #[derivative(PartialEq = "ignore")]
     pub(super) stats: CoiStats,
 
     // removed fields go below this line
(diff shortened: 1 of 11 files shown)


xayn-ai/src/coi/point.rs
```diff
@@ -121,6 +152,15 @@ pub(crate) trait CoiPoint {
...
 
+    fn key_phrases(&self) -> &[KeyPhrase] {
+        &[]
+    }
+
+    fn swap_key_phrases(&mut self, candidates: Vec<KeyPhrase>) -> Vec<KeyPhrase> {
+        #![allow(unused_variables)]
+        Vec::new()
+    }
+
```
rustonaut, line 163: I think it's better to not make this  default impls. 

Given that the new method explicitly accepts `key_phrases` this feels rather wrong.

xayn-ai/src/embedding/utils.rs
```diff
@@ -57,23 +57,32 @@ where
 ///
 /// # Panics
 /// Panics if the vectors don't consist solely of real values or their shapes don't match.
+///
+/// # Safety
+/// Only use this with [`TrustedLen`] iterators. The bound can't be named currently, because the
+/// trait is nightly-gated. [`ExactSizeIterator`] isn't a feasible replacement, because it isn't
```
rustonaut, line 63: > [`ExactSizeIterator`] isn't a feasible replacement,

It's also not "trusted", i.e. it must be safe for it to lie about it's length.

xayn-ai/src/embedding/utils.rs
```diff
@@ -57,23 +57,32 @@ where
...
+    let size = iter.size_hint().0;
 
     let norms = iter.clone().map(|a| l2_norm(a.view())).collect::<Vec<_>>();
     let mut similarities = Array2::ones((size, size));
     for (i, a) in iter.clone().enumerate() {
         if norms[i] != 0. {
             for (j, b) in iter.clone().enumerate().skip(i + 1) {
                 if norms[j] != 0. {
-                    similarities[[i, j]] = a.dot(b) / norms[i] / norms[j];
+                    similarities[[i, j]] = a.dot(&b) / norms[i] / norms[j];
```
rustonaut, line 85: As this uses bounds checks we don't really need to  make the function `unsafe` a bad `size` will just panic.

xayn-ai/src/coi/point.rs
janpetschexain: i agree. runtime checking this via using `unimplemented!()` in the default impl would be one way, i think having it compile time checked is even better, so i have split up the `CoiPoint` into additional traits in the same fashion as `CoiPointMerge` before. since the same argument applies for the stats as well, i've splitted out `CoiPointKeyPhrases` and `CoiPointStats` and only implemented them for the `PositiveCoi` and `NegativeCoi`. if we want to avoid the impls for `NegativeCoi` as well, we would have to split the methods of the `CoiSystem`, which we might consider later on but which is out of scope for now.

xayn-ai/src/embedding/utils.rs
janpetschexain: updated the docs

xayn-ai/src/embedding/utils.rs
janpetschexain: removed the `unsafe`, updated the docs and calls

xayn-ai/src/embedding/utils.rs
```diff
@@ -57,23 +57,31 @@ where
...
+/// Only use this with [`TrustedLen`] iterators, otherwise indexing may panic. The bound can't be
+/// named currently, because the trait is nightly-gated. [`ExactSizeIterator`] isn't a feasible and
+/// trusted replacement, for example it isn't implemented for [`Chain`]ed iterators.
+///
+/// [`TrustedLen`]: std::iter::TrustedLen
+/// [`ExactSizeIterator`]: std::iter::ExactSizeIterator
+/// [`Chain`]: std::iter::Chain
 #[allow(dead_code)]
-pub fn pairwise_cosine_similarity<'a, I, S>(iter: I) -> Array2<f32>
+pub fn pairwise_cosine_similarity<I, S>(iter: I) -> Array2<f32>
```
acrrd, line 69: Seems that we have lot's of complexity for a minor optimization. 
Can't just we do something like:
```rust
iter.clone().enumerate().cartesian_product(iter.enumerate())
  .filter(|((i, _), (j, _))| i > j)
  .map(/* similarity here */)
  .collect()
``` 
?
We are already iterating more then necessary by doing `skip`  anyway.
The other option would be to take a `Vec<ArrayBase>` or a ref to it, so we can allocate the correct amount and iterate only the needed times.

xayn-ai/src/coi/point.rs
```diff
@@ -180,24 +225,52 @@ impl CoiPoint for PositiveCoi_v0_1_0 {
...
 
     coi_point_default_impls! {}
+}
+
+impl CoiPointKeyPhrases for PositiveCoi {
+    fn key_phrases(&self) -> &[KeyPhrase] {
+        self.key_phrases.as_slice()
+    }
 
+    fn swap_key_phrases(&mut self, mut candidates: Vec<KeyPhrase>) -> Vec<KeyPhrase> {
```
acrrd, line 265: Why we need this? and why we are returning the candidates, we are loosing `key_phrases`

xayn-ai/src/coi/point.rs
```diff
@@ -50,6 +53,26 @@ impl Default for CoiStats {
...
+    pub(super) point: Embedding,
+}
+
+impl KeyPhrase {
+    pub(super) fn is_valid(&self) -> bool {
+        !self.words.is_empty() && self.point.iter().copied().all(f32::is_finite)
+    }
+
+    pub(super) fn is_unique(key_phrases: &[KeyPhrase]) -> bool {
+        key_phrases
```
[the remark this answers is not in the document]
janpetschexain: no we can't, as explained in the pr description:

> the uniqueness invariant can't be enforced for example by a HashSet because KeyPhrase can't implement Eq + Hash and also not by sort & dedup because KeyPhrase can't implement PartialOrd/Ord

this is because of the limitation of the underlying `ArrayBase`.

xayn-ai/src/embedding/utils.rs
janpetschexain: i just timed it, the cartesian product approach is about 30% faster and the vector approach is about 30% slower than the skip approach, so i'm going to change it to the former.

xayn-ai/src/coi/point.rs
janpetschexain: we may change the function signature later on, for now it is implemented this way to be able to iterate over the owned key phrases in the follow up pr.

not sure what you mean by losing the `key_phrases`? there is a memory swap of `key_phrases` and `candidates`, meaning that the previous `key_phrases` are actually returned and the `candidates` become the new key phrases.

xayn-ai/src/coi/point.rs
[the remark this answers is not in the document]
acrrd: if we can assume that two `KeyPhrase` with the same `words` then also the embedding is the same we can implement `Eq` and `Hash` ignoring the embedding.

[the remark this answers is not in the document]
janpetschexain: unfortunately we can't assume that. for example `"key phrase"` and `"Key Phrase"` may or may not produce different embeddings, e.g. in cased bert models. so we would have to rely on `SMBert` never changing, otherwise we have errors in here without noticing.

xayn-ai/src/coi/point.rs
```diff
@@ -180,24 +230,48 @@ impl CoiPoint for PositiveCoi_v0_1_0 {
...
 
 impl CoiPoint for PositiveCoi {
-    fn new(id: CoiId, point: Embedding, viewed: Option<Duration>) -> Self {
+    fn new(
+        id: CoiId,
+        point: Embedding,
+        key_phrases: Vec<KeyPhrase>,
+        viewed: Option<Duration>,
+    ) -> Self {
+        debug_assert!(key_phrases.iter().all(KeyPhrase::is_valid));
```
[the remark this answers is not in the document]
janpetschexain: removed it since the `KeyPhrase` constructor ensures this now.

xayn-ai/src/coi/point.rs
```diff
@@ -50,6 +53,26 @@ impl Default for CoiStats {
     }
 }
 
+#[derive(Clone, Deserialize, PartialEq, Serialize)]
+#[cfg_attr(test, derive(Debug))]
```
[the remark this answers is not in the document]
janpetschexain: we have to keep in mind that this can increase the binary size, because we have to move some dev-dependencies to dependencies (e.g. like `derivative`).

xayn-ai/src/coi/point.rs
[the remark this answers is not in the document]
janpetschexain: changed as discussed.

acrrd (approved):

janpetschexain:
bors merge

PR #353, opened 2021-12-20
[Merged by Bors] - TY-2208 update key phrases [2]

janpetschexain:
**References**

- [TY-2208]
- #348
- #361
- #371

**Summary**

- make `SMBert` available to `CoiSystem` & replace the smbert closure
- mock `SMBert` in the coi tests & adjust reranker builder
- add more key phrase selection tests

[TY-2208]: https://xainag.atlassian.net/browse/TY-2208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #355, opened 2021-12-22
[Merged by Bors] - TY-2209 top key phrases

janpetschexain:
**References**

- [TY-2209]
- #348
- #353
- #361

**Summary**

- rename `compute_weights()` to `compute_relevances()` to match the source
- move the `horizon` from `compute_relevances()` to the coi `Configuration` and make `max_key_phrases` a constant of the config
- add selection for the top key phrases of a set of cois:
  - move key phrase storing from the `CoiPoint` to the `CoiSystem`
  - update coi relevance and key phrase relevance computation
  - add tests and update all affected tests
- `assert_approx_eq!`: add `BTreeSet` support
- key phrases will not be addressed in merging of cois as this requires major redesign, we agreed on moving this to the new engine as synchronization is not really needed for now

[TY-2209]: https://xainag.atlassian.net/browse/TY-2209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test-utils/src/approx_eq.rs
+++ b/test-utils/src/approx_eq.rs
@@ -1,4 +1,4 @@
-use std::iter;
+use std::{collections::BTreeSet, iter};
 
 use float_cmp::ApproxEq;
 use ndarray::{ArrayBase, Data, Dimension, IntoDimension, Ix};
@@ -64,12 +64,12 @@ macro_rules! assert_approx_eq {
     ($t:ty, $left:expr, $right:expr, epsilon = $epsilon:expr, ulps = $ulps:expr $(,)?) => {{
         let epsilon = $epsilon;
         let ulps = $ulps;
-        let left = $left;
-        let right = $right;
+        let left = &$left;
+        let right = &$right;
         let mut left_iter =
-            $crate::ApproxEqIter::<$t>::indexed_iter_logical_order(&left, Vec::new());
+            $crate::ApproxEqIter::<$t>::indexed_iter_logical_order(left, Vec::new());
         let mut right_iter =
-            $crate::ApproxEqIter::<$t>::indexed_iter_logical_order(&right, Vec::new());
+            $crate::ApproxEqIter::<$t>::indexed_iter_logical_order(right, Vec::new());
         loop {
             match (left_iter.next(), right_iter.next()) {
                 (std::option::Option::Some((lidx, lv)), std::option::Option::Some((ridx, rv))) => {
@@ -135,7 +135,10 @@ where
 macro_rules! impl_approx_eq_iter {
     ($($t:ty),+ $(,)?) => {
         $(
-            impl<'a> ApproxEqIter<'a, $t> for &'a $t {
+            impl<'a> ApproxEqIter<'a, $t> for &'a $t
+            where
+                $t: Copy,
+            {
                 fn indexed_iter_logical_order(
                     self,
                     index_prefix: Vec<Ix>,
@@ -226,6 +229,22 @@ macro_rules! impl_approx_eq_iter {
                         let mut index_prefix = index_prefix.clone();
                         index_prefix.extend(idx.into_dimension().as_array_view().iter());
                         (index_prefix, *el)
+                    }))
+                }
+            }
+
+            impl<'a, T> ApproxEqIter<'a, $t> for &'a BTreeSet<T>
+            where
+                &'a T: ApproxEqIter<'a, $t>,
+            {
+                fn indexed_iter_logical_order(
+                    self,
+                    index_prefix: Vec<Ix>,
+                ) -> Box<dyn Iterator<Item = (Vec<Ix>, $t)> + 'a> {
+                    Box::new(self.iter().enumerate().flat_map(move |(idx, el)| {
+                        let mut index_prefix = index_prefix.clone();
+                        index_prefix.push(idx);
+                        el.indexed_iter_logical_order(index_prefix)
                     }))
                 }
             }
--- a/xayn-ai/src/coi/config.rs
+++ b/xayn-ai/src/coi/config.rs
@@ -1,4 +1,6 @@
-use std::num::NonZeroUsize;
+use std::{num::NonZeroUsize, time::Duration};
+
+use crate::utils::SECONDS_PER_DAY;
 
 #[derive(Clone, Copy)]
 pub(crate) struct Configuration {
@@ -8,6 +10,9 @@ pub(crate) struct Configuration {
     pub threshold: f32,
     /// The positive number of neighbors for the k-nearest-neighbors distance.
     pub neighbors: NonZeroUsize,
+    /// The time since the last view after which a coi becomes irrelevant.
+    #[allow(dead_code)]
+    pub horizon: Duration,
     /// The maximum number of key phrases picked during the coi key phrase selection. A coi may have
     /// more key phrases than this, eg because of merging.
     #[allow(dead_code)]
@@ -23,6 +28,7 @@ impl Default for Configuration {
             shift_factor: 0.1,
             threshold: 12.0,
             neighbors: NonZeroUsize::new(4).unwrap(),
+            horizon: Duration::from_secs(SECONDS_PER_DAY as u64 * 30),
             max_key_phrases: 3,
             gamma: 0.9,
         }
--- a/xayn-ai/src/coi/config.rs
+++ b/xayn-ai/src/coi/config.rs
@@ -2,7 +2,7 @@ use std::{num::NonZeroUsize, time::Duration};
 
 use crate::utils::SECONDS_PER_DAY;
 
-#[derive(Clone, Copy)]
+#[derive(Clone)]
 pub(crate) struct Configuration {
     /// The shift factor by how much a Coi is shifted towards a new point.
     pub shift_factor: f32,
@@ -20,6 +20,8 @@ pub(crate) struct Configuration {
     /// The weighting between coi and pairwise candidate similarites in the key phrase selection.
     #[allow(dead_code)]
     pub gamma: f32,
+    /// The penalty for less relevant key phrases in the top key phrase selection.
+    pub penalty: Vec<f32>,
 }
 
 impl Default for Configuration {
@@ -31,6 +33,7 @@ impl Default for Configuration {
             horizon: Duration::from_secs(SECONDS_PER_DAY as u64 * 30),
             max_key_phrases: 3,
             gamma: 0.9,
+            penalty: vec![1., 0.75, 0.66],
         }
     }
 }
--- a/xayn-ai/src/coi/config.rs
+++ b/xayn-ai/src/coi/config.rs
@@ -20,7 +20,9 @@ pub(crate) struct Configuration {
     /// The weighting between coi and pairwise candidate similarites in the key phrase selection.
     #[allow(dead_code)]
     pub gamma: f32,
-    /// The penalty for less relevant key phrases in the top key phrase selection.
+    /// The penalty for less relevant key phrases of a coi in increasing order (ie. lowest penalty
+    /// for the most relevant key phrase first and highest penalty for the least relevant key phrase
+    /// last).
     pub penalty: Vec<f32>,
 }
 
--- a/xayn-ai/src/coi/key_phrase.rs
+++ b/xayn-ai/src/coi/key_phrase.rs
@@ -1,21 +1,16 @@
-use std::{
-    borrow::{Borrow, Cow},
-    collections::BTreeSet,
-    convert::identity,
-    iter::once,
-};
+use std::{borrow::Borrow, collections::BTreeSet, convert::identity, iter::once, time::Duration};
 
 use derivative::Derivative;
-use lazy_static::lazy_static;
 use ndarray::{s, Array1, Array2, ArrayBase, Axis, Data, Ix, Ix2};
 use serde::{Deserialize, Serialize};
 
 use crate::{
     coi::{
-        config::Configuration,
         point::{CoiPoint, NegativeCoi, PositiveCoi},
-        stats::{compute_relevances, CoiPointStats},
+        relevance::{Relevance, RelevanceMaps},
+        stats::CoiPointStats,
         CoiError,
+        CoiId,
     },
     embedding::utils::{pairwise_cosine_similarity, ArcEmbedding, Embedding},
     error::Error,
@@ -27,13 +22,6 @@ pub(crate) struct KeyPhrase {
     words: String,
     #[derivative(Ord = "ignore", PartialEq = "ignore", PartialOrd = "ignore")]
     point: ArcEmbedding,
-    #[derivative(Ord = "ignore", PartialEq = "ignore", PartialOrd = "ignore")]
-    relevance: f32,
-}
-
-lazy_static! {
-    // TODO: temporary workaround, remove once positive and negative cois have been split properly
-    static ref EMPTY_KEY_PHRASES: BTreeSet<KeyPhrase> = BTreeSet::new();
 }
 
 impl KeyPhrase {
@@ -43,7 +31,6 @@ impl KeyPhrase {
     ) -> Result<Self, CoiError> {
         let words = words.into();
         let point = point.into();
-        let relevance = 0.;
 
         if words.is_empty() || point.is_empty() {
             return Err(CoiError::EmptyKeyPhrase);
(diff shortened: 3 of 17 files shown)


xayn-ai/src/coi/config.rs
```diff
@@ -7,11 +9,18 @@ pub(crate) struct Configuration {
...
     /// The weighting between coi and pairwise candidate similarites in the key phrase selection.
     pub gamma: f32,
+    /// The penalty for less relevant key phrases in the top key phrase selection.
+    pub penalty: [f32; Self::MAX_KEY_PHRASES],
+}
+
+impl Configuration {
+    /// The maximum number of key phrases picked during the coi key phrase selection. A coi may have
+    /// more key phrases than this, eg because of merging.
+    pub const MAX_KEY_PHRASES: usize = 3;
```
acrrd, line 23: Can't we then use `penalty.len()` as `max_key_phrases`? This with `penalty: [f32; Self::MAX_KEY_PHRASES]` constraints the configuration more then needed.

xayn-ai/src/coi/system.rs
```diff
@@ -186,17 +186,11 @@ impl CoiSystem {
...
-    /// irrelevant.
-    #[allow(dead_code)]
-    fn compute_weights<CP: CoiPoint + CoiPointStats>(
-        &self,
-        cois: &[CP],
-        horizon: Duration,
-    ) -> Vec<f32> {
+    /// The relevance of each coi is computed from its view count and view time relative to the
+    /// other cois. It's an unnormalized score from the interval `[0, ∞)`.
+    fn compute_relevances<CP: CoiPoint + CoiPointStats>(&self, cois: &[CP]) -> Vec<f32> {
```
acrrd, line 193: Also here I think is better if we have a core that does not depends on self and the use it with the configuration we have.

xayn-ai/src/coi/system.rs
```diff
@@ -294,7 +288,7 @@ impl CoiSystem {
             .collect::<BTreeSet<_>>();
 
         let len = coi.key_phrases().len() + candidates.len();
-        let max_key_phrases = self.config.max_key_phrases.min(len);
+        let max_key_phrases = Configuration::MAX_KEY_PHRASES.min(len);
```
acrrd, line 291: The problem with this is that we cannot test with different values.

xayn-ai/src/coi/system.rs
```diff
@@ -382,6 +376,48 @@ impl CoiSystem {
...
+            .iter_mut()
+            .zip(relevances)
+            .map(|(coi, relevance)| {
+                self.select_key_phrases(coi, &[]);
+                let mut key_phrases = coi
+                    .key_phrases()
+                    .iter()
+                    .map(|key_phrase| (key_phrase.words(), key_phrase.relevance()))
+                    .collect::<Vec<_>>();
+                key_phrases.sort_unstable_by(|(_, this), (_, other)| {
```
acrrd, line 402: We can store them already with the order we want in `select_key_phrases`.

xayn-ai/src/coi/config.rs
```diff
@@ -7,11 +9,18 @@ pub(crate) struct Configuration {
...
     pub neighbors: NonZeroUsize,
-    /// The maximum number of key phrases picked during the coi key phrase selection. A coi may have
-    /// more key phrases than this, eg because of merging.
-    pub max_key_phrases: usize,
+    /// The time since the last view after which a coi becomes irrelevant.
+    pub horizon: Duration,
     /// The weighting between coi and pairwise candidate similarites in the key phrase selection.
     pub gamma: f32,
+    /// The penalty for less relevant key phrases in the top key phrase selection.
+    pub penalty: [f32; Self::MAX_KEY_PHRASES],
```
acrrd, line 17: We may want to check that the passed configuration for penalty contains finite numbers.

xayn-ai/src/coi/system.rs
```diff
@@ -382,6 +376,48 @@ impl CoiSystem {
...
+                    this.partial_cmp(other).unwrap(/* relevance is never nan */)
+                });
+                key_phrases.into_iter().rev().zip(self.config.penalty).map(
+                    move |((key_phrase, _), penalty)| {
+                        (key_phrase, (relevance * penalty).max(f32::MIN))
```
acrrd, line 407: Why with need the max with min?

xayn-ai/src/coi/system.rs
```diff
@@ -382,6 +376,48 @@ impl CoiSystem {
...
+                    },
+                )
+            })
+            .flatten()
+            .collect::<Vec<_>>();
+        relevant_key_phrases.sort_unstable_by(|(_, this), (_, other)| this.partial_cmp(other).unwrap(/* penalized relevance is never nan */));
```
acrrd, line 413: I slightly prefer to do `this.partial_cmp(other).unwrap().reverse()` here then `rev` on the iterator.

xayn-ai/src/coi/system.rs
janpetschexain: i guess the same argument applies as in the other pr: as the whole module is structured this way, this method is as well. as we agreed to change it in another pr, i'd like to leave it as is for now.

xayn-ai/src/coi/system.rs
janpetschexain: i'll change it to a const generic to allow for testing with different values.

xayn-ai/src/coi/config.rs
janpetschexain: we don't have any checks for any value of the configuration yet. i agree that it makes sense to check values before accepting them in the config, i think it makes sense to do this in a separate pr, wdyt?

xayn-ai/src/coi/system.rs
janpetschexain: to guarantee that the resulting value is clamped into finite floats (`f32::clamp()` won't do this).

xayn-ai/src/coi/system.rs
janpetschexain: the new draining algorithm to select the top key phrases is implemented now. two things to note which are different from the original suggestion: instead of
```rust
coiToScore: HashMap<CoiId, Vec<f32>;
phrasesByScore: BTreeMap<f32, Vec<(String, CoiId)>>
```
we have
```rust
coi_to_relevance: HashMap<CoiId, BTreeSet<Relevance>>,
relevance_to_key_phrase: BTreeMap<(Relevance, CoiId), Vec<KeyPhrase>>,
```
where `Relevance` wraps an `f32`. the wrapper is required because the keys of the map must implement `Ord`. the id in the second map is moved from the values to the keys, as this makes iterating & construction & draining more convenient, the desired sorting is still the same.

xayn-ai/src/coi/system.rs
```diff
@@ -866,101 +960,203 @@ mod tests {
...
+        assert_approx_eq!(f32, system.coi_to_relevance[&cois[0].id], [0.]);
+        assert_approx_eq!(f32, system.coi_to_relevance[&cois[1].id], [0.]);
+        assert!(system.relevance_to_key_phrase.is_empty());
+    }
+
+    fn dedup<const N: usize>(array: [f32; N]) -> Array1<f32> {
+        assert!(IntoIterator::into_iter(array).all(f32::is_finite));
+        IntoIterator::into_iter(array)
+            .map(Relevance)
+            .collect::<BTreeSet<_>>()
```
acrrd, line 994: Can't we use an hashset? Or we also want them ordered?

xayn-ai/src/coi/system.rs
```diff
@@ -866,101 +960,203 @@ mod tests {
...
-        let horizon = Duration::from_secs_f32(SECONDS_PER_DAY);
-        let weights = CoiSystem::default().compute_weights(&cois, horizon);
-        assert_approx_eq!(f32, weights, [0.5, 0.6666667, 0.8333333], epsilon = 0.00001);
+        system.compute_relevances(&cois);
+        assert_eq!(system.coi_to_relevance.len(), 3);
+        assert_approx_eq!(
+            f32,
+            system.coi_to_relevance[&cois[0].id],
+            0.5 * dedup(system.config.penalty),
+            epsilon = 0.00001,
```
acrrd, line 1016: Seems that in the rest of the code `ulps` is used more, can we used that instead?

xayn-ai/src/coi/system.rs
```diff
@@ -866,101 +960,203 @@ mod tests {
...
+        };
+        let mut system = CoiSystem::new(config, SMBert::default());
         let mut cois = create_pos_cois(&[[1., 2., 3.], [4., 5., 6.], [7., 8., 9.]]);
         cois[0].stats.last_view -= Duration::from_secs_f32(0.5 * SECONDS_PER_DAY);
         cois[1].stats.last_view -= Duration::from_secs_f32(1.5 * SECONDS_PER_DAY);
         cois[2].stats.last_view -= Duration::from_secs_f32(2.5 * SECONDS_PER_DAY);
-        let horizon = Duration::from_secs_f32(2. * SECONDS_PER_DAY);
-        let weights = CoiSystem::default().compute_weights(&cois, horizon);
+        system.compute_relevances(&cois);
+        assert_eq!(system.coi_to_relevance.len(), 3);
```
acrrd, line 1078:
```suggestion
        assert_eq!(system.coi_to_relevance.len(), cois.len());
```
seems more rubust

test-utils/src/approx_eq.rs
```diff
@@ -64,12 +64,12 @@ macro_rules! assert_approx_eq {
     ($t:ty, $left:expr, $right:expr, epsilon = $epsilon:expr, ulps = $ulps:expr $(,)?) => {{
         let epsilon = $epsilon;
         let ulps = $ulps;
-        let left = $left;
-        let right = $right;
+        let left = &$left;
+        let right = &$right;
         let mut left_iter =
-            $crate::ApproxEqIter::<$t>::indexed_iter_logical_order(&left, Vec::new());
+            $crate::ApproxEqIter::<$t>::indexed_iter_logical_order(left, Vec::new());
```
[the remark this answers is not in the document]
janpetschexain: it is necessary, without it the macro takes the arguments by value instead of by reference (because of the prior assignment some lines above), which may require to clone them before calling the macro. the change aligns the `assert_approx_eq!` macro with the behavior of other macros like `assert_eq!`.

xayn-ai/src/coi/system.rs
```diff
@@ -40,16 +42,38 @@ pub(crate) enum CoiSystemError {
     NoMatchingDocuments,
 }
 
+/// A finite relevance score.
+#[derive(Clone, Copy, Debug, Default, Into, PartialEq, PartialOrd)]
+pub struct Relevance(f32);
+
+impl Eq for Relevance {
+    // never nan by construction
```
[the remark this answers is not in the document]
janpetschexain: no one outside of the module can construct it because the field is private. within the module we guarantee that the invariant holds. this is no different if we had an explicit constructor function.

xayn-ai/src/coi/system.rs
```diff
@@ -40,16 +42,38 @@ pub(crate) enum CoiSystemError {
...
+        self.partial_cmp(other).unwrap(/* never nan by construction */)
+    }
+}
+
 #[cfg_attr(test, derive(Default))]
 pub(crate) struct CoiSystem {
     config: Configuration<3>,
     smbert: SMBert,
+    coi_to_relevance: HashMap<CoiId, BTreeSet<Relevance>>,
+    relevance_to_key_phrase: BTreeMap<(Relevance, CoiId), Vec<KeyPhrase>>,
```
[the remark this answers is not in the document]
janpetschexain: the `BTreeSet` is mainly there for sorting (uniqueness is not a concern), we could use a `Vec` instead, but then we would need to either sort this vec when combining both maps or we would need to guarantee that the vec is sorted when constructed.

xayn-ai/src/coi/key_phrase.rs
```diff
@@ -83,26 +53,14 @@ impl Borrow<str> for KeyPhrase {
...
+        self.words.eq(other)
     }
 }
 
-impl CoiPointKeyPhrases for NegativeCoi {
-    fn key_phrases(&self) -> &BTreeSet<KeyPhrase> {
-        &EMPTY_KEY_PHRASES
+impl PartialEq<KeyPhrase> for &str {
+    fn eq(&self, other: &KeyPhrase) -> bool {
+        other.words.eq(self)
```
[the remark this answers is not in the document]
janpetschexain: we use `KeyPhrase == &str` for the tests, it's probably unexpected if someone later wants to use `&str == KeyPhrase` and this isn't implemented. as long as it's not used, it shouldn't be compiled into the binary as well. shall i still remove it?

xayn-ai/src/coi/system.rs
```diff
@@ -203,57 +224,101 @@ impl CoiSystem {
...
+                .penalty
+                .iter()
+                .map(|&penalty| {
+                    Relevance(
+                        (f32::from(new_relevance) * penalty)
+                            .max(f32::MIN)
+                            .min(f32::MAX),
+                    )
+                })
+                .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
janpetschexain: imo it would look weird to split the body of the loop into another function, because it depends on `counts` & `times` which don't make much sense out of context for a single coi. also testing can be done without too much effort for the whole function. shall i still split it out?

xayn-ai/src/coi/config.rs
```diff
@@ -2,7 +2,7 @@ use std::{num::NonZeroUsize, time::Duration};
 
 use crate::utils::SECONDS_PER_DAY;
 
-pub(crate) struct Configuration {
+pub(crate) struct Configuration<const MAX_KEY_PHRASES: usize> {
```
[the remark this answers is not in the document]
janpetschexain: i guess we then have to change the `penalty` from an array to a vec. i'd like to postpone this and handle it in the config validation task.

xayn-ai/src/coi/system.rs
```diff
@@ -207,59 +222,103 @@ impl CoiSystem {
...
-        candidates: &[String],
-    ) {
-        /// Filters the unique candidates wrt the existing key phrases.
-        fn unique_candidates<CP>(
-            coi: &CP,
+    ///
+    /// The relevances in the maps of this system are replaced by the key phrase relevances.
+    fn select_key_phrases<CP: CoiPoint>(&mut self, coi: &mut CP, candidates: &[String]) {
+        /// Unites the key phrases and candidates of the coi.
+        fn unique(
```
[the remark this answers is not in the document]
janpetschexain: i'll split it further.

xayn-ai/src/coi/system.rs
```diff
@@ -399,48 +443,92 @@ impl CoiSystem {
...
+        let (relevance_to_coi, key_phrases) = self
+            .relevance_to_key_phrase
+            .iter()
+            .map(|((relevance, coi_id), key_phrases)| {
+                key_phrases
+                    .iter()
+                    .map(move |key_phrase| ((*relevance, *coi_id), key_phrase.clone()))
+            })
+            .flatten()
+            .rev()
```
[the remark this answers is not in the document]
janpetschexain: not sure what you mean by before `iter()`, do you mean before the `flatten()` or the `map()`? it might only change the order of key phrases with equal relevance, but we don't guarantee that order anyways, so it should be ok to change it. i'll test it and see if i missed sth.

xayn-ai/src/coi/system.rs
janpetschexain: correct, we want them ordered to compare in the tests without additional sorting.

xayn-ai/src/coi/system.rs
janpetschexain: this depends a lot on what we want to test here. the `ulps` is a measure for a relative error whereas the `epsilon` is a measure for an absolute error. for example both of the following will pass even though the absolute difference in the second case is huge:
```rust
assert_approx_eq!(f32, 9999999e-7, 9999998e-7, epsilon = 0., ulps = 2);
assert_approx_eq!(f32, 9999999e+10, 9999998e+10, epsilon = 0., ulps = 2);
```
in our case we know that the error is roughly about `1e-5` to `1-e6` due to the nature of the compared values and we can express this with the `epsilon` to have a more precise test.

xayn-ai/src/coi/system.rs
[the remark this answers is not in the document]
janpetschexain: > Can we move this to another struct that implement the function to update a coi and to get the top k?

done

xayn-ai/src/coi/key_phrase.rs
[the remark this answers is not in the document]
acrrd: not sure, or we add a comment to say that is for that reason or we remove it, we add it later or we always write 
`KeyPhrase == &str`

xayn-ai/src/coi/system.rs
acrrd: ok make sense. Thanks!

xayn-ai/src/coi/relevance.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+/// Sorted maps from cois to relevances to key phrases.
+#[derive(Debug, Default)]
+pub(crate) struct Relevances {
+    coi_to_relevance: HashMap<CoiId, BTreeSet<Relevance>>,
+    relevance_to_key_phrase: BTreeMap<(Relevance, CoiId), Vec<KeyPhrase>>,
+}
+
+impl Relevances {
```
acrrd, line 46: All `pub` here can be `pub(super)`.

acrrd (approved):

xayn-ai/src/coi/key_phrase.rs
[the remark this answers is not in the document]
janpetschexain: removed it for now

janpetschexain:
bors merge

janpetschexain:
bors merge

janpetschexain:
bors merge

PR #367, opened 2022-01-07
[Merged by Bors] - TY-2190 add compute_smbert method

Robert-Steiner:
Ticket:

- [TY-2190]

Summary:
- adds `compute_smbert` method
- We decided to go with the signature `fn compute_smbert(&self, sequence: &str) -> Result<Embedding, Error>`. That way we don’t depend on the layout of `Document`.

[TY-2190]: https://xainag.atlassian.net/browse/TY-2190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/xayn-ai/src/ranker/mod.rs
+++ b/xayn-ai/src/ranker/mod.rs
@@ -3,6 +3,7 @@ use rubert::SMBert;
 
 use crate::{
     coi::{point::UserInterests, CoiSystem},
+    embedding::utils::Embedding,
     error::Error,
 };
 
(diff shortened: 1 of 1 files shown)


xayn-ai/src/ranker/mod.rs
```diff
@@ -3,6 +3,7 @@ use rubert::SMBert;
 
 use crate::{
     coi::{point::UserInterests, CoiSystem},
+    embedding::utils::Embedding,
```
janpetschexain, line 6: depending on if/how many times we need to clone the embedding in the new ranker, we could also use the new `ArcEmbedding` instead of `Embedding`.

janpetschexain (approved):

xayn-ai/src/ranker/mod.rs
Robert-Steiner: Good point! I will leave it as it is for now because I'm not sure yet how often we will need to clone the embedding but I will keep it in mind.

Robert-Steiner:
bors merge

PR #372, opened 2022-01-11
[Merged by Bors] - TY-2232 rank method

Robert-Steiner:
Ticket:

[TY-2232]

Summary:

- adds the `rank` method to the `Ranker` struct

[TY-2232]: https://xainag.atlassian.net/browse/TY-2232?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


xayn-ai/src/ranker/context.rs
```diff
@@ -0,0 +1,117 @@
+use crate::{data::document_data::CoiComponent, ranker::utils::Id};
+
+/// The context used to calculate a document's score.
+/// https://xainag.atlassian.net/wiki/spaces/M2D/pages/770670607/Production+AI+Workflow#3.2-Context-calculations.
```
janpetschexain, line 4: i'm not sure if we should put private links (private in the sense of access permissions, not in the sense of classified) in public code?

xayn-ai/src/ranker/context.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// Average positive distance.
+    pos_avg: f32,
+    /// Maximum negative distance.
+    neg_max: f32,
+}
+
+impl Context {
+    pub(crate) fn from_cois(cois: &[(Id, CoiComponent)]) -> Self {
+        let cois_len = cois.len() as f32;
+        let pos_avg = cois.iter().map(|(_, coi)| coi.pos_distance).sum::<f32>() / cois_len;
```
janpetschexain, line 16: if cois is empty, then the pos_avg will be nan, is that ok for the context or should it be handled?

xayn-ai/src/ranker/context.rs
```diff
@@ -0,0 +1,117 @@
...
+        let neg_max = cois
+            .iter()
+            .map(|(_, coi)| coi.neg_distance)
+            .fold(f32::MIN, f32::max); // NOTE f32::max considers NaN as smallest value
+
+        Self { pos_avg, neg_max }
+    }
+
+    /// Calculates score from given positive distance and negative distance.
+    pub(crate) fn calculate_score(&self, pos: f32, neg: f32) -> f32 {
```
janpetschexain, line 26: since pos and neg are assumed to be distances, should we check that they are actually >= 0?

xayn-ai/src/ranker/mod.rs
```diff
@@ -1,34 +1,55 @@
...
+}
+
 /// The Ranker.
 pub(crate) struct Ranker {
-    #[allow(dead_code)]
-    coi: CoiSystem,
+    /// SMBert system.
     smbert: SMBert,
+    /// CoI configuration.
+    coi_config: Configuration,
```
janpetschexain, line 34: i still need the coi system for the feedback loop, but i agree that it makes sense to have the config directly in the ranker. maybe in another pr: we could move the `Configuration` from `coi::config::` to `ranker::config` and use it as the general config for the ranker, not only for the coi. or we could create a config for the ranker which holds a coi config next to possibly other configs. wdyt?

xayn-ai/src/ranker/mod.rs
```diff
@@ -43,4 +64,143 @@ impl Ranker {
...
+    /// Ranks the given documents based on the learned user interests.
+    ///
+    /// # Errors
+    ///
+    /// Fails if no user interests are known.
+    pub(crate) fn rank(&self, documents: &mut [Document]) -> Result<(), Error> {
+        rank(
+            documents,
+            &self.user_interests,
+            self.coi_config.neighbors.get(),
```
janpetschexain, line 77: just as a heads up, once #375 is merged this will change to
```suggestion
            self.coi_config.neighbors(),
```

xayn-ai/src/ranker/mod.rs
```diff
@@ -43,4 +64,143 @@ impl Ranker {
...
+    user_interests: &UserInterests,
+    neighbors: usize,
+) -> Result<Vec<(DocumentId, CoiComponent)>, Error> {
+    documents
+        .iter()
+        .map(|document| {
+            let coi =
+                compute_coi_for_embedding(&document.smbert_embedding, user_interests, neighbors)
+                    .ok_or(RankerError::NoUserInterests)?;
+            Ok::<(DocumentId, CoiComponent), Error>((document.id, coi))
```
janpetschexain, line 111: the function signature is sufficient for the type inference
```suggestion
            Ok((document.id, coi))
```

xayn-ai/src/ranker/mod.rs
```diff
@@ -43,4 +64,143 @@ impl Ranker {
...
+) -> HashMap<&DocumentId, f32> {
+    let context = Context::from_cois(cois_for_docs);
+    let mut scores = HashMap::new();
+    cois_for_docs.iter().for_each(|(id, coi)| {
+        scores.insert(
+            id,
+            context.calculate_score(coi.pos_distance, coi.neg_distance),
+        );
+    });
+    scores
```
janpetschexain, line 127: this could be collected directly:
```suggestion
    cois_for_docs
        .iter()
        .map(|(id, coi)| {
            (
                id,
                context.calculate_score(coi.pos_distance, coi.neg_distance),
            )
        })
        .collect()
```

xayn-ai/src/ranker/mod.rs
Robert-Steiner: interesting, it didn't work before, but the code was also a bit different

xayn-ai/src/ranker/mod.rs
Robert-Steiner: nice!

xayn-ai/src/ranker/context.rs
Robert-Steiner: I guess it would be fine. cois can only be empty when `documents` is empty and when we have no documents we don't call `calculate_score`. We can add a check at the the beginning of `rank` and return early if the `documents` is empty wdyt?

xayn-ai/src/ranker/context.rs
Robert-Steiner: I agree, but my hope is that at some point it will be publicly accessible 🙏. We also have the issue with the jira tickets

xayn-ai/src/ranker/mod.rs
Robert-Steiner: I agree that is a good idea 👍

xayn-ai/src/ranker/context.rs
Robert-Steiner: I'm not sure if it's worth to add a runtime check here and what we should do if one of them is < 0. We can add to the docstring that both values need to be >= 0.

janpetschexain: adding some documentation is good. and we can also add some debug assertions.

Robert-Steiner: done

janpetschexain (approved):

Robert-Steiner:
bors merge

PR #373, opened 2022-01-12
Add tests to `ranker::Context`

x3ro:


xayn-ai/src/ranker/context.rs
```diff
@@ -19,13 +21,82 @@ impl Context {
         Self { pos_avg, neg_max }
     }
 
-    /// Calculates context value from given LTR score, positive distance, negative distance and similarity.
-    pub(crate) fn calculate(&self, pos: f32, neg: f32) -> f32 {
+    /// Calculates context value from given positive distance, negative distance.
```
Robert-Steiner, line 24:
```suggestion
    /// Calculates score from given positive distance and negative distance.
```

xayn-ai/src/ranker/context.rs
```diff
@@ -1,5 +1,7 @@
 use crate::{data::document_data::CoiComponent, ranker::util::Id};
 
+/// Ranking context calculation according to
+/// [the spec](https://xainag.atlassian.net/wiki/spaces/M2D/pages/770670607/Production+AI+Workflow#3.2-Context-calculations.)
```
Robert-Steiner, line 4:
```suggestion
/// The context used to calculate a document's score.
```
I like the idea of ​​linking the document, but it describes the old context calculation and that could be a bit confusing.

xayn-ai/src/ranker/context.rs
```diff
@@ -19,13 +21,82 @@ impl Context {
...
+
+        let cxt = calc.calculate_score(8., 7.);
+        assert_approx_eq!(f32, cxt, ((1. / 3.) + (1. / 2.)) / 2.);
+    }
+
+    #[test]
+    fn test_from_cois() {
+        let cois = vec![
+            (
+                Id::from_rand(),
```
Robert-Steiner, line 73: we can also use `Id::from_u128(0)` because for the context calculation it doesn't matter if the id is unique

Robert-Steiner (commented):
Looks good 👍 only found some minor things

x3ro:
Re: the documentation link: Personally I'd probably keep it around, since it'd be easier to understand the code with outdated info than it was with no info at all. I don't feel strongly about it though.

Robert-Steiner (approved):
> Re: the documentation link: Personally I'd probably keep it around, since it'd be easier to understand the code with outdated info than it was with no info at all. I don't feel strongly about it though.

we could add:
```
/// https://xainag.atlassian.net/wiki/spaces/M2D/pages/770670607/Production+AI+Workflow#3.2-Context-calculations.
/// outlines the calculation of positive and negative distance factor.
```

x3ro:
Would you put this on `Context` or `calculate_score`?

PR #376, opened 2022-01-13
[Merged by Bors] - TY-2283 - New design for builder pattern

x3ro:
[TY-2283]

Unfortunately builders aren't trivial to encode in Rust's type system, which is why we currently have the problem with lots of generics in the builders type signature.

We could use [derive-builder](https://crates.io/crates/derive_builder), which does all checks (completeness) at runtime. Alternatively there is [rust-typed-builder](https://github.com/idanarye/rust-typed-builder), which works similar to the builder approach in use currently, which has the downside of generating a plethora of generics and is not very customizable. 

As a result, I'm proposing to get rid of the "fluent interface" for required parameters, so that they have to be specified up-front. For example:

Before:

```rust
let smbert = SMBertBuilder::from_files("vocab.txt", "model.onnx")?
    .with_accents(false)
    .with_lowercase(true)
    .with_token_size(64)?
    .with_pooling(FirstPooler)
    .build()?;
```

After:

```rust
let config = Configuration::from_files("vocab.txt", "model.onnx")?
                    .with_accents(false)
                    .with_lowercase(true)
                    .with_pooling(FirstPooler)
                    .with_token_size(64)?;

let smbert = Pipeline::from(config);
```

This avoids most of the generic parameters except for the pooler, for which there is no common super-type at the moment (i.e. NonePooler, AveragePooler, etc don't share a trait).

[TY-2283]: https://xainag.atlassian.net/browse/TY-2283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


xayn-ai/src/reranker/public.rs
```diff
@@ -141,108 +140,59 @@ impl<SV, SM, QAV, QAM, DM> Builder<SV, SM, QAV, QAM, DM> {
...
         let domain = self.domain.build()?;
-        let context = Context;
-        let analytics = AnalyticsSystemImpl;
 
         super::Reranker::new(Systems {
-            database,
+            database: self.database,
             smbert,
-            qambert,
+            qambert: rubert::Pipeline::from(qambert_config)?,
```
Robert-Steiner, line 191: same here
```suggestion
            qambert: QAMBert::from(qambert_config)?,
```

xayn-ai/src/reranker/public.rs
```diff
@@ -141,108 +140,59 @@ impl<SV, SM, QAV, QAM, DM> Builder<SV, SM, QAV, QAM, DM> {
...
-        let qambert = self
-            .qambert
+        let smbert_config = self
+            .smbert_config
+            .with_token_size(52)?
+            .with_accents(false)
+            .with_lowercase(true)
+            .with_pooling(AveragePooler);
+
+        let smbert_pipeline = rubert::Pipeline::from(smbert_config)?;
```
Robert-Steiner, line 175: what about
```suggestion
        let smbert_pipeline = rubert::SMBert::from(smbert_config)?;
```

or we use `As`
```rust
use rubert::{AveragePooler, QAMBert, QAMBertConfig, SMBert, SMBertConfig};

use crate::embedding::smbert::SMBert As ArcSMBert;

let smbert_pipeline = SMBert::from(smbert_config)?;
```

xayn-ai/src/ranker/public.rs
```diff
@@ -65,107 +58,35 @@ impl<SV, SM, KV, KM> Builder<SV, SM, KV, KM> {
...
+
+        let kpe_config = self
+            .kpe_config
             .with_token_size(150)?
             .with_accents(false)
-            .with_lowercase(false)
-            .build()?;
+            .with_lowercase(false);
 
+        let kpe = kpe::Pipeline::from(kpe_config)?;
```
Robert-Steiner, line 86: maybe
```rust
use kpe::Pipeline as KPE;

let kpe = KPE::from(kpe_config)?;
```
to make it look similar to SMBert

xayn-ai/src/lib.rs
```diff
@@ -34,6 +34,11 @@ pub use crate::{
     },
 };
 
+// We need to re-export these, since they encapsulate the arguments
+// required for pipeline construction, and are passed to builders.
+pub use kpe::Configuration as KpeConfiguration;
```
Robert-Steiner, line 39: `Configuration` is currently not used
```suggestion
```

xayn-ai/src/embedding/qambert.rs
```diff
@@ -90,15 +90,15 @@ mod tests {
...
             .with_token_size(90)
             .unwrap()
             .with_accents(false)
             .with_lowercase(true)
-            .with_pooling(AveragePooler)
-            .build()
-            .unwrap()
+            .with_pooling(AveragePooler);
+
+        rubert::Pipeline::from(config).unwrap()
```
Robert-Steiner, line 101:
```suggestion
        QAMBert::from(config).unwrap()
```

rubert/src/pipeline.rs
```diff
@@ -32,6 +33,27 @@ pub enum PipelineError {
...
+    K: BertModel,
+{
+    pub fn from(config: Configuration<K, P>) -> Result<Self, PipelineError> {
+        let tokenizer = Tokenizer::new(
+            config.vocab,
+            config.accents,
+            config.lowercase,
+            config.token_size,
+        )?;
+        let model = Model::new(config.model, config.token_size)?;
```
Robert-Steiner, line 47: We can also map those errors to those we had in the builder 
```rust
    /// Failed to build the tokenizer: {0}
    Tokenizer(#[from] TokenizerError),
    /// Failed to build the model: {0}
    Model(#[from] ModelError),
```
this way we also keep the old error messages in `xayn-ai-ffi-c/src/reranker/ai.rs`

Robert-Steiner (commented):
Looks good so far 👍
I will take another look after the rebase

xayn-ai/src/ranker/public.rs
x3ro: Mmh, for me this only makes it harder to understand what's going on. Or in other words, why introduce an additional alias for a type, which the reader of the code then has to resolve?

rubert/src/pipeline.rs
x3ro: We could, but since the pipeline can also return errors from its `run` methods, they would be slightly off in one of the scenarios in any case.

xayn-ai/src/reranker/public.rs
x3ro: See comment above. I don't see the benefit, from the code reader's perspective, of introducing an alias here. Can you elaborate why this was decided in the past?

xayn-ai/src/ranker/public.rs
Robert-Steiner: I personally think it is easier to read than 

```rust
kpe::Pipeline::from(kpe_config)
rubert::Pipeline::from(smbert_config);
rubert::Pipeline::from(qambert_config);
```

rubert/src/pipeline.rs
Robert-Steiner: What I meant was adding them as additional variants

```rust
pub enum PipelineError {
    /// Failed to run the tokenizer: {0}
    Tokenizer(#[from] TokenizerError),
    /// Failed to run the model: {0}
    Model(#[from] ModelError),
    /// Failed to run the pooler: {0}
    Pooler(#[from] PoolerError),
    /// Failed to build the tokenizer: {0}
    TokenizerBuild(#[source] TokenizerError),
    /// Failed to build the model: {0}
    ModelBuild(#[source] ModelError),
}

impl<K, P> Pipeline<K, P>
where
    K: BertModel,
{
    pub fn from(config: Configuration<K, P>) -> Result<Self, PipelineError> {
        let tokenizer = Tokenizer::new(
            config.vocab,
            config.accents,
            config.lowercase,
            config.token_size,
        )
        .map_err(PipelineError::TokenizerBuild)?;
        let model =
            Model::new(config.model, config.token_size).map_err(PipelineError::ModelBuild)?;

        Ok(Pipeline {
            tokenizer,
            model,
            pooler: config.pooler,
        })
    }
}
```

xayn-ai/src/ranker/public.rs
x3ro: As discussed, we'll leave it as is for now and think of better ways to name it than `Pipeline` in the future.

x3ro:
I think I've covered everything :)

kpe/src/configuration.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// Defaults to `None`. The actual returned number of ranked key phrases might be less than
+    /// indicated by the threshold depending on the upper count for the key phrases.
+    ///
+    /// # Errors
+    /// Fails if `score` is given and not finite.
+    pub fn with_key_phrase_min_score(
+        mut self,
+        score: Option<f32>,
+    ) -> Result<Self, ConfigurationError> {
+        if score.is_none() || score.map(f32::is_finite).unwrap_or_default() {
```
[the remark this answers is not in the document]
x3ro: This didn't compile, I went with `score.map(...).is_some()` instead.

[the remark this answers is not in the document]
acrrd: It actualy is `score.filter(|v| f32::is_finite(*v)).is_some()`.
With `map` and `is_some` does not what we want. `Some(NaN)` will map to `Some(false)` and the check will pass so or we go back to `map`/`unwrap` or we use filter.

kpe/examples/kpe.rs
```diff
@@ -0,0 +1,21 @@
...
+use test_utils::kpe::*;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let config = Configuration::from_files(
+        vocab().unwrap(),
+        bert().unwrap(),
+        cnn().unwrap(),
+        classifier().unwrap(),
+    )?
+    .with_token_size(128)?;
```
Robert-Steiner, line 12:
```suggestion
    let config = Configuration::from_files(vocab()?, bert()?, cnn()?, classifier()?)?
        .with_token_size(128)?;
```

a bit shorter

kpe/src/pipeline.rs
```diff
@@ -26,9 +28,32 @@ pub enum PipelineError {
     Tokenizer(#[from] TokenizerError),
     /// Failed to run the model: {0}
     Model(#[from] ModelError),
+    /// Failed to load binary parameters from a file: {0}
+    BinParams(#[from] LoadingBinParamsFailed),
```
Robert-Steiner, line 32: Can you please also add the two errors here and map them in the `from` method 

```
    /// Failed to build the tokenizer: {0}
    TokenizerBuild(#[source] TokenizerError),
    /// Failed to build the model: {0}
    ModelBuild(#[source] ModelError),
```

it will make easier to debug

Robert-Steiner (commented):
two small things but the rest looks good to me

x3ro:
@acrrd @Robert-Steiner ready to merge?

Robert-Steiner (approved):
🚀

x3ro:
bors merge

x3ro:
bors merge

PR #377, opened 2022-01-14
[Merged by Bors] - TY-2230 feedback loop [0]

janpetschexain:
**References**

- [TY-2230]

**Summary**

preparation for the feedback loop impl:
- split `CoiPointConstructor` (only test convenience) from `CoiPoint`
- move coi point shifting into `CoiPoint`
- split update of positive/negative cois
- remove no-ops in negative coi updates
- remove unused `CoiPointStats`/`CoiPointKeyPhrases` traits

[TY-2230]: https://xainag.atlassian.net/browse/TY-2230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/xayn-ai/src/coi/system.rs
+++ b/xayn-ai/src/coi/system.rs
@@ -7,10 +7,15 @@ use uuid::Uuid;
 use crate::{
     coi::{
         config::Configuration,
-        key_phrase::CoiPointKeyPhrases,
-        point::{find_closest_coi, find_closest_coi_mut, UserInterests},
+        point::{
+            find_closest_coi,
+            find_closest_coi_mut,
+            CoiPoint,
+            NegativeCoi,
+            PositiveCoi,
+            UserInterests,
+        },
         relevance::Relevances,
-        stats::CoiPointStats,
         utils::{classify_documents_based_on_user_feedback, collect_matching_documents},
         CoiId,
     },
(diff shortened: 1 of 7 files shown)


xayn-ai/src/coi/system.rs
```diff
@@ -111,34 +116,19 @@ pub(crate) fn compute_coi(
...
-) -> Vec<CP> {
+) -> Vec<PositiveCoi> {
     match find_closest_coi_mut(embedding, &mut cois, config.neighbors()) {
         Some((coi, distance)) if distance < config.threshold() => {
-            coi.set_point(shift_coi_point(
-                embedding,
-                coi.point(),
-                config.shift_factor(),
-            ));
-            coi.set_id(Uuid::new_v4().into());
```
Robert-Steiner, line 141: This is no longer needed? (just to be sure)

Robert-Steiner (approved):
splitting `update_cois` into positive and negative makes the coi code much cleaner

xayn-ai/src/coi/system.rs
janpetschexain: correct, i also double checked this. it was convenient for the merge/sync implementation (which we don't use in the near future), but it's very inconvenient for any centralized data structure mapping coi ids to some data (which we will use for many things in the new engine).

janpetschexain:
bors merge

PR #386, opened 2022-01-18
[Merged by Bors] - TY-2292 add last_view to NegativeCoi

Robert-Steiner:
Ticket: 
- [TY-2292]

Summary:
- adds `last_view` field to `NegativeCoi` struct
- `last_view` is used later in the context calculation

[TY-2292]: https://xainag.atlassian.net/browse/TY-2292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/xayn-ai/src/coi/point.rs
+++ b/xayn-ai/src/coi/point.rs
@@ -1,4 +1,4 @@
-use std::time::Duration;
+use std::time::{Duration, SystemTime};
 
 use derivative::Derivative;
 use serde::{Deserialize, Serialize};
@@ -6,7 +6,7 @@ use serde::{Deserialize, Serialize};
 use crate::{
     coi::{stats::CoiStats, CoiId},
     embedding::utils::{l2_distance, Embedding},
-    utils::nan_safe_f32_cmp,
+    utils::{nan_safe_f32_cmp, system_time_now},
 };
 
 #[obake::versioned]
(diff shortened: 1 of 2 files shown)


xayn-ai/src/coi/point.rs
```diff
@@ -71,17 +72,21 @@ impl From<PositiveCoi_v0_2_0> for PositiveCoi {
...
 }
 
-#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[derive(Clone, Debug, Derivative, Deserialize, Serialize)]
+#[derivative(PartialEq)]
 pub(crate) struct NegativeCoi {
     pub(super) id: CoiId,
     pub(super) point: Embedding,
+    #[derivative(PartialEq = "ignore")]
+    pub(super) last_view: SystemTime,
```
rustonaut, line 81: Ignoring fields for equality comparison is generally a bad idea (it e.g. can lead so surprising results when, e.g. de-duplicating a list). But I guess it might be fine here?

janpetschexain: the statistics of a coi can be viewed as meta data about the coi which doesn't change what constitutes the coi itself. in that sense it shouldn't affect the comparison, this is also how we handle other stats in the cois.

Robert-Steiner: Thanks for the answer, I wanted to ask you about it today

Robert-Steiner:
bors merge

PR #387, opened 2022-01-18
[Merged by Bors] - TY-2240 weighted ranking

Robert-Steiner:
Ticket:

- [TY-2240]

part of [TY-2138]

- replaces the current context calcualtion with the new weighted context calcualtion
- [python counterpart](https://github.com/xaynetwork/xayn_ai_research/blob/5c97c8f96511bcc69e8b2bc74853346afe37d3f4/xayn_ai/backend/rerankers/coi/coi_reranker_3_factor.py#L107)

[TY-2240]: https://xainag.atlassian.net/browse/TY-2240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[TY-2138]: https://xainag.atlassian.net/browse/TY-2138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/xayn-ai/src/ranker/config.rs
+++ b/xayn-ai/src/ranker/config.rs
@@ -15,8 +15,8 @@ pub(crate) struct Configuration {
     max_key_phrases: usize,
     gamma: f32,
     penalty: Vec<f32>,
-    min_positive_cois: u32,
-    min_negative_cois: u32,
+    min_positive_cois: usize,
+    min_negative_cois: usize,
 }
 
 /// Potential errors of the ranker configuration.
@@ -165,7 +165,7 @@ impl Configuration {
     }
 
     /// The minimum number of positive cois required for the context calculation.
-    pub(crate) fn min_positive_cois(&self) -> u32 {
+    pub(crate) fn min_positive_cois(&self) -> usize {
         self.min_positive_cois
     }
 
@@ -174,7 +174,7 @@ impl Configuration {
     /// # Errors
     /// Fails if the minimum number is zero.
     #[allow(dead_code)]
-    pub(crate) fn with_min_positive_cois(self, min_positive_cois: u32) -> Result<Self, Error> {
+    pub(crate) fn with_min_positive_cois(self, min_positive_cois: usize) -> Result<Self, Error> {
         if min_positive_cois > 0 {
             Ok(Self {
                 min_positive_cois,
@@ -186,7 +186,7 @@ impl Configuration {
     }
 
     /// The minimum number of negative cois required for the context calculation.
-    pub(crate) fn min_negative_cois(&self) -> u32 {
+    pub(crate) fn min_negative_cois(&self) -> usize {
         self.min_negative_cois
     }
 
@@ -195,7 +195,7 @@ impl Configuration {
     /// # Errors
     /// Fails if the minimum number is zero.
     #[allow(dead_code)]
-    pub(crate) fn with_min_negative_cois(self, min_negative_cois: u32) -> Result<Self, Error> {
+    pub(crate) fn with_min_negative_cois(self, min_negative_cois: usize) -> Result<Self, Error> {
         if min_negative_cois > 0 {
             Ok(Self {
                 min_negative_cois,
--- a/xayn-ai/src/ranker/config.rs
+++ b/xayn-ai/src/ranker/config.rs
@@ -15,6 +15,8 @@ pub(crate) struct Configuration {
     max_key_phrases: usize,
     gamma: f32,
     penalty: Vec<f32>,
+    min_positive_cois: u32,
+    min_negative_cois: u32,
 }
 
 /// Potential errors of the ranker configuration.
@@ -31,6 +33,10 @@ pub(crate) enum Error {
     Gamma,
     /// Invalid coi penalty, expected non-empty, finite and sorted values
     Penalty,
+    /// Invalid minimum number of positive cois, expected positive value
+    MinPositiveCois,
+    /// Invalid minimum number of negative cois, expected positive value
+    MinNegativeCois,
 }
 
 impl Configuration {
@@ -102,7 +108,7 @@ impl Configuration {
         Self { horizon, ..self }
     }
 
-    /// The weighting between coi and pairwise candidate similarites in the key phrase selection.
+    /// The weighting between coi and pairwise candidate similarities in the key phrase selection.
     pub(crate) fn gamma(&self) -> f32 {
         self.gamma
     }
@@ -157,6 +163,48 @@ impl Configuration {
     pub(crate) fn max_key_phrases(&self) -> usize {
         self.penalty.len()
     }
+
+    /// The minimum number of positive cois required for the context calculation
+    pub(crate) fn min_positive_cois(&self) -> u32 {
+        self.min_positive_cois
+    }
+
+    /// Sets the minimum number of positive cois.
+    ///
+    /// # Errors
+    /// Fails if the minimum number is zero.
+    #[allow(dead_code)]
+    pub(crate) fn with_min_positive_cois(self, min_positive_cois: u32) -> Result<Self, Error> {
+        if min_positive_cois > 0 {
+            Ok(Self {
+                min_positive_cois,
+                ..self
+            })
+        } else {
+            Err(Error::MinPositiveCois)
+        }
+    }
+
+    /// The minimum number of negative cois required for the context calculation.
+    pub(crate) fn min_negative_cois(&self) -> u32 {
+        self.min_negative_cois
+    }
+
+    /// Sets the minimum number of negative cois.
+    ///
+    /// # Errors
+    /// Fails if the minimum number is zero.
+    #[allow(dead_code)]
+    pub(crate) fn with_min_negative_cois(self, min_negative_cois: u32) -> Result<Self, Error> {
+        if min_negative_cois > 0 {
+            Ok(Self {
+                min_negative_cois,
+                ..self
+            })
+        } else {
+            Err(Error::MinNegativeCois)
+        }
+    }
 }
 
 impl Default for Configuration {
@@ -169,6 +217,8 @@ impl Default for Configuration {
             max_key_phrases: 3,
             gamma: 0.9,
             penalty: vec![1., 0.75, 0.66],
+            min_positive_cois: 2,
+            min_negative_cois: 2,
         }
     }
 }
--- a/xayn-ai/src/ranker/config.rs
+++ b/xayn-ai/src/ranker/config.rs
@@ -164,7 +164,7 @@ impl Configuration {
         self.penalty.len()
     }
 
-    /// The minimum number of positive cois required for the context calculation
+    /// The minimum number of positive cois required for the context calculation.
     pub(crate) fn min_positive_cois(&self) -> u32 {
         self.min_positive_cois
     }
--- a/xayn-ai/src/ranker/context.rs
+++ b/xayn-ai/src/ranker/context.rs
@@ -60,6 +60,9 @@ fn find_closest_cois(
     .into()
 }
 
+///# Panics
+///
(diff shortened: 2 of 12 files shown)


xayn-ai/src/ranker/config.rs
```diff
@@ -157,6 +163,48 @@ impl Configuration {
     pub(crate) fn max_key_phrases(&self) -> usize {
         self.penalty.len()
     }
+
+    /// The minimum number of positive cois required for the context calculation
```
janpetschexain, line 167:
```suggestion
    /// The minimum number of positive cois required for the context calculation.
```

xayn-ai/src/ranker/context.rs
```diff
@@ -1,119 +1,231 @@
...
-            .fold(f32::MIN, f32::max); // NOTE f32::max considers NaN as smallest value
+/// Helper struct for [`find_closest_cois`].
+struct ClosestCois {
+    /// The ID of the positive centre of interest
+    pos_id: CoiId,
+    /// Distance from the positive centre of interest
+    pos_distance: f32,
+    pos_last_view: SystemTime,
+
+    /// Distance from the negative centre of interest
```
janpetschexain, line 42: to reduce ambiguity
```suggestion
    /// The ID of the closest positive centre of interest
    pos_id: CoiId,
    /// Distance from the closest positive centre of interest
    pos_distance: f32,
    pos_last_view: SystemTime,

    /// Distance from the closest negative centre of interest
```

xayn-ai/src/ranker/context.rs
```diff
@@ -1,119 +1,231 @@
...
+}
 
-        Self { pos_avg, neg_max }
+fn find_closest_cois(
+    embedding: &Embedding,
+    user_interests: &UserInterests,
+    neighbors: usize,
+) -> Option<ClosestCois> {
+    let (pos_coi, pos_distance) = find_closest_coi(&user_interests.positive, embedding, neighbors)?;
+    let (neg_coi, neg_distance) = find_closest_coi(&user_interests.negative, embedding, neighbors)?;
```
janpetschexain, line 53: the `find_closest_coi()` computation is different to the pink code, i assume this will be changed later when the l2 distances are replaced with the cosine similarities?

xayn-ai/src/ranker/context.rs
```diff
@@ -1,119 +1,231 @@
...
 
-    /// Calculates score from given positive distance and negative distance.
-    /// Both positive and negative distance must be >= 0.
-    pub(crate) fn calculate_score(&self, pos: f32, neg: f32) -> f32 {
-        debug_assert!(pos >= 0. && neg >= 0.);
-        let frac_pos = (self.pos_avg > 0.)
-            .then(|| (1. + pos / self.pos_avg).recip())
-            .unwrap_or(1.);
-        let frac_neg = (1. + (self.neg_max - neg)).recip();
+    let score = cois.pos_distance * pos_decay + pos_coi_relevance - cois.neg_distance * neg_decay;
```
janpetschexain, line 83: we could just return it without an additional binding
```suggestion
    Ok(cois.pos_distance * pos_decay + pos_coi_relevance - cois.neg_distance * neg_decay)
```

janpetschexain, line 83: do we need to make sure that the score is finite? (it can never be nan if all components aren't nan, but just in case we can still account for this in the same step)
```suggestion
    let score = (cois.pos_distance * pos_decay + pos_coi_relevance - cois.neg_distance * neg_decay).max(f32::MIN).min(f32::MAX);
```

xayn-ai/src/ranker/config.rs
```diff
@@ -15,6 +15,8 @@ pub(crate) struct Configuration {
     max_key_phrases: usize,
     gamma: f32,
     penalty: Vec<f32>,
+    min_positive_cois: u32,
+    min_negative_cois: u32,
```
janpetschexain, line 19: is there a specific reason that those two are `u32`? otherwise we could make them `usize` as usual.

xayn-ai/src/ranker/context.rs
```diff
@@ -1,119 +1,231 @@
...
+
+    let neg_cois = user_interests.negative.len() as u32;
+    if neg_cois < min_negative_cois {
+        return Err(Error::NotEnoughNegativeCois {
+            expected: min_negative_cois,
+            has: neg_cois,
+        });
+    }
+    Ok(())
+}
```
janpetschexain, line 109: do we need the detailed error information? otherwise we could simplify this to
```suggestion
fn has_enough_cois(
    user_interests: &UserInterests,
    min_positive_cois: u32,
    min_negative_cois: u32,
) -> bool {
    user_interests.positive.len() >= min_positive_cois && user_interests.negative.len() >= min_negative_cois
}
```
and return a simplified error at the call site like `Error::NotEnoughCois`.

xayn-ai/src/ranker/context.rs
```diff
@@ -1,119 +1,231 @@
...
+
+/// Computes the score for all documents based on the given information.
+///
+/// <https://xainag.atlassian.net/wiki/spaces/M2D/pages/2240708609/Discovery+engine+workflow#The-weighting-of-the-CoI>
+/// outlines parts of the score calculation.
+///
+/// # Errors
+/// Fails if the required number of positive or negative cois is not present or
+/// if coi relevance is not present in the relevances map.
```
janpetschexain, line 118: this case should never happen because `compute_score_for_embedding()` guarantees that the relevances are freshly computed.

xayn-ai/src/ranker/context.rs
```diff
@@ -1,119 +1,231 @@
...
+    horizon: Duration,
+    now: SystemTime,
+) -> Result<f32, Error> {
+    let cois = find_closest_cois(embedding, user_interests, neighbors)
+        .ok_or(Error::FailedToFindTheClosestCois)?;
+
+    let pos_decay = compute_coi_decay_factor(horizon, now, cois.pos_last_view);
+    let neg_decay = compute_coi_decay_factor(horizon, now, cois.neg_last_view);
+
+    relevances.compute_relevances(&user_interests.positive, horizon);
```
janpetschexain, line 79: at this point we probably should make `now: SystemTime` also an argument for the `compute_relevances()` function such that all time-dependent computations use the same `now`.

xayn-ai/src/ranker/context.rs
```diff
@@ -1,119 +1,231 @@
...
+        user_interests,
+        config.min_positive_cois(),
+        config.min_negative_cois(),
+    )?;
+
+    let now = system_time_now();
+    documents
+        .iter()
+        .map(|document| {
+            let score = compute_score_for_embedding(
```
janpetschexain, line 135: currently we compute the coi relevances in each iteration of the loop again, but they will be the same everytime because the relevant input arguments don't change. i'd suggest to pull out `compute_relevances()` from `compute_score_for_embedding()`. then the latter can also take `relevances` just by `&` instead of `&mut`.
```suggestion
    relevances.compute_relevances(&user_interests.positive, horizon);
    documents
        .iter()
        .map(|document| {
            let score = compute_score_for_embedding(
```

xayn-ai/src/ranker/context.rs
```diff
@@ -1,119 +1,231 @@
...
+        let score = compute_score_for_embedding(
+            &embedding,
+            &user_interests,
+            &mut relevances,
+            1,
+            horizon,
+            now,
+        )
+        .unwrap();
+        // 1.1185127 * 0.99999934 + 0.73094887 - 0 * 100 = 12
```
janpetschexain, line 208: what does the `= 12` refer to?

xayn-ai/src/ranker/mod.rs
```diff
@@ -81,9 +80,14 @@ impl Ranker {
...
     ///
-    /// Fails if no user interests are known.
-    pub(crate) fn rank(&self, documents: &mut [impl Document]) -> Result<(), Error> {
-        rank(documents, &self.user_interests, self.config.neighbors())
+    /// Fails if the scores of the documents cannot be computed.
+    pub(crate) fn rank(&mut self, documents: &mut [impl Document]) -> Result<(), Error> {
+        rank(
+            documents,
+            &self.user_interests,
+            &mut self.coi.relevances_mut(),
```
janpetschexain, line 88:
```suggestion
            self.coi.relevances_mut(),
```

xayn-ai/src/ranker/context.rs
Robert-Steiner: That was the result of the old calculation. I forgot to update it

Robert-Steiner: That was the result of the old calculation. I forgot to update it

xayn-ai/src/ranker/context.rs
Robert-Steiner: Yes, later we will use cosine similarities but other than that, did you find anything else that is different in the calculation?

xayn-ai/src/ranker/config.rs
Robert-Steiner: No specific reason. I will replace it with `usize`

xayn-ai/src/ranker/context.rs
Robert-Steiner: Good catch! 
what would be the best way to get the relevance from the map?

```rust
    let pos_coi_relevance = relevances
        .relevance_for_coi(&cois.pos_id).unwrap(/* we calculate relevance for all positive cois one line above */);
```

shall we add an error type for that and call unwrap on `compute_score_for_embedding` or do you think it is enough to add a note to the doc string of `compute_score_for_embedding`?

xayn-ai/src/ranker/context.rs
janpetschexain: no, the computation looks the same (of course except for places where we will replace the l2 distance with cosine similarity). we have to double check the [order of sorting](https://github.com/xaynetwork/xayn_ai/blob/TY-2240-weighted-ranking/xayn-ai/src/ranker/mod.rs#L143-L148) once the change is done.

xayn-ai/src/ranker/context.rs
Robert-Steiner: I've added the infomation because I thought it might help with debugging, but yeah it might be too much. We can add this back if we later find it helpful.

xayn-ai/src/ranker/context.rs
janpetschexain: an additional error type is probably too much, i would just unwrap within `compute_score_for_embedding()` and add a `# Panics` section to its documentation. at the call site in `compute_score_for_docs()` we could add a short comment that calling it will not panic because we just computed all required relevances.

xayn-ai/src/ranker/context.rs
Robert-Steiner: Yes I think we can add this

xayn-ai/src/ranker/context.rs
Robert-Steiner: done

xayn-ai/src/ranker/context.rs
Robert-Steiner: i removed that part

xayn-ai/src/ranker/context.rs
Robert-Steiner: done

xayn-ai/src/ranker/config.rs
Robert-Steiner: done

janpetschexain (approved):
lgtm 👍

Robert-Steiner:
bors merge

Robert-Steiner:
bors merge

PR #388, opened 2022-01-19
TY-2291 Use cosine similarity

x3ro:
[TY-2291]


xayn-ai/src/coi/system.rs
```diff
@@ -81,35 +81,34 @@ impl systems::CoiSystem for CoiSystem {
...
-    neighbors: usize,
 ) -> Option<CoiComponent> {
-    let (coi, pos_distance) = find_closest_coi(embedding, &user_interests.positive, neighbors)?;
-    let neg_distance = match find_closest_coi(embedding, &user_interests.negative, neighbors) {
+    let (coi, pos_similarity) = find_closest_coi(embedding, &user_interests.positive)?;
+    let neg_similarity = match find_closest_coi(embedding, &user_interests.negative) {
         Some((_, dis)) => dis,
-        None => f32::MAX,
+        // TODO: Is this correct? This was previously f32::MAX, but given that we
+        //       now have a similarity score instead of a distance, I'd assume it should be 0.
```
janpetschexain, line 93: i'd say it should be `-1` as this is the lowest possible similarity, equivalently to the previous highest distance.

xayn-ai/src/coi/system.rs
```diff
@@ -180,8 +184,8 @@ fn update_negative_coi(
     embedding: &Embedding,
     config: &Configuration,
 ) -> Vec<NegativeCoi> {
-    match find_closest_coi_mut(embedding, &mut cois, config.neighbors()) {
-        Some((coi, distance)) if distance < config.threshold() => {
+    match find_closest_coi_mut(embedding, &mut cois) {
+        Some((coi, distance)) if distance >= config.threshold() => {
```
janpetschexain, line 188:
```suggestion
        Some((coi, similarity)) if similarity >= config.threshold() => {
```

xayn-ai/src/ranker/mod.rs
```diff
@@ -78,26 +80,22 @@ impl Ranker {
...
 
-    let cois_for_docs = compute_cois_for_docs(documents, user_interests, neighbors)?;
+    let cois_for_docs = compute_cois_for_docs(documents, user_interests)?;
     let score_for_docs = compute_score_for_docs(cois_for_docs.as_slice());
 
     documents.sort_unstable_by(|a, b| {
         nan_safe_f32_cmp(
-            score_for_docs.get(&b.id).unwrap(),
             score_for_docs.get(&a.id).unwrap(),
+            score_for_docs.get(&b.id).unwrap(),
```
janpetschexain, line 98: this change reverses the order of the sorted results, i'm not sure if this is intended, that's why i wanted to double check?

xayn-ai/src/ranker/context.rs
```diff
@@ -13,10 +13,10 @@ pub(crate) struct Context {
 impl Context {
     pub(crate) fn from_cois(cois: &[(Id, CoiComponent)]) -> Self {
         let cois_len = cois.len() as f32;
-        let pos_avg = cois.iter().map(|(_, coi)| coi.pos_distance).sum::<f32>() / cois_len;
+        let pos_avg = cois.iter().map(|(_, coi)| coi.pos_similarity).sum::<f32>() / cois_len;
```
janpetschexain, line 16: the `calculate_score()` below was meant for distances, i guess it has to be updated to use similarities as well now? if the formulas stay the same, we probably still have to update the default values in the unwrappings.

janpetschexain (commented):
> What to do with the failing reranker tests?

i think we have to decide if it's worth the effort to update tests for systems which wont be used in the new engine (and just update the tests which matter later on, ignore the rest). wdyt @acrrd?

xayn-ai/src/embedding/utils.rs
```diff
@@ -55,7 +55,9 @@ where
 
 /// Computes the pairwise cosine similarities of vectors.
 ///
-/// Zero vectors are chosen to be similar to all other vectors, i.e. a similarity of 1.
+/// * For positive vectors, cosine similarity is bounded in [0, 1]
```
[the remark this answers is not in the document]
x3ro: I mean vectors in positive space, i.e. all vectors where all values are strictly >= 0. I'll adjust the wording.

xayn-ai/src/ranker/mod.rs
x3ro: That's intended. The previous metric was a distance (smaller is better) whereas now we have a similarity (larger is better), which is why I think we need the order flipped.

x3ro:
From your comments I gather that the COIs and embedding we expect will not be constrained to positive space, is that correct? I got this impression because e.g. all of the tests throughout the entire system which I've touched never deal with anything negative. If that's the case I should go over the PR once more, and also add tests for vectors in negative space.

xayn-ai/src/embedding/utils.rs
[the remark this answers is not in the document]
janpetschexain: thx, i see. i think the term is "non-negative cone of the vector space". but a description of what is meant is also fine.

xayn-ai/src/ranker/mod.rs
janpetschexain: the metric changed, but the `calculate_score()` function which takes the changed metric and is used to compute the `score_for_docs` was not changed. this is then probably related to https://github.com/xaynetwork/xayn_ai/pull/388#discussion_r789038633.

janpetschexain:
> From your comments I gather that the COIs and embedding we expect will not be constrained to positive space, is that correct? I got this impression because e.g. all of the tests throughout the entire system which I've touched never deal with anything negative. If that's the case I should go over the PR once more, and also add tests for vectors in negative space.

in general the embeddings are from the whole real-valued vector space, for example the smbert embeddings are from the whole IR^128. it's quite unfortunate that the tests give the impression that the embedding space is somewhat limited to a half or a cone of the whole space, but that's not the case. (the only assumption we can make is that the values of the embeddings are somewhat close to zero and very far from positive/negative infinity, because of normalization/regularization performed within the embedding model, but this doesn't help much.)

xayn-ai/src/embedding/utils.rs
```diff
@@ -67,11 +69,10 @@ where
 /// [`TrustedLen`]: std::iter::TrustedLen
 /// [`ExactSizeIterator`]: std::iter::ExactSizeIterator
 /// [`Chain`]: std::iter::Chain
-pub fn pairwise_cosine_similarity<I, S>(iter: I) -> Array2<f32>
+pub fn pairwise_cosine_similarity<'a, I>(iter: I) -> Array2<f32>
 where
-    I: IntoIterator<Item = ArrayBase<S, Ix1>>,
+    I: IntoIterator<Item = ArrayView1<'a, f32>>,
```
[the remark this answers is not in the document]
x3ro: I made this change to simplify the signatures. I deemed this reasonable because, on any owned representation, calling `.view()` will convert it to `ArrawView`.

x3ro:
@janpetschexain @acrrd Okay so regarding the reranker tests. I played around with them some more, but I don't think they're a trivial fix (i.e. just changing the expeced outcomes), though that takes care of _most_ of the failures. Even if we changed those expected outcomes, however, I feel like the tests wouldn't really be worth something if we also verified that the expected outcomes we've added are correct.

My understanding is that the reranker is not going to be used, so should we potentially ignore it for now, as @janpetschexain suggested above?

xayn-ai/src/ranker/mod.rs
x3ro: Good point. Is [this](https://github.com/xaynetwork/xayn_ai_research/blob/c93a032884e40196e41b5fdb6ac610937c2d0234/xayn_ai/backend/rerankers/coi/coi_reranker_3_factor.py#L1410) (minus the time decay) how we would want to implement this?

So essentially just

```rust
pos_similarity - neg_similarity
```

🤔

xayn-ai/src/coi/point.rs
```diff
@@ -183,78 +183,50 @@ impl From<UserInterests_v0_2_0> for UserInterests {
...
-                .take(neighbors)
-                .zip(distances.iter().take(neighbors).rev())
-                .map(|((_, distance), (_, reversed))| distance * (reversed / total))
-                .sum()
-        })
-        .unwrap_or_default();
-
-    Some((index, distance))
+
+    similarities.sort_by(|(_, this), (_, other)| other.partial_cmp(this).unwrap());
```
janpetschexain, line 201: i'd prefer to use `reverse()` to make it obvious that the sorting order is reversered
```suggestion
    similarities.sort_by(|(_, this), (_, other)| this.partial_cmp(other).unwrap().reverse());
```

xayn-ai/src/coi/system.rs
```diff
@@ -139,35 +142,32 @@ impl systems::CoiSystem for CoiSystem {
 /// Assigns a CoI for the given embedding.
 ///
 /// Returns `None` if no CoI could be found otherwise it returns the Id of
-/// the CoL along with the positive and negative distance. The negative distance
-/// will be [`f32::MAX`], if no negative coi could be found.
+/// the CoL along with the positive and negative similarity.
```
janpetschexain, line 145: typo
```suggestion
/// the CoI along with the positive and negative similarity.
```

xayn-ai/src/coi/system.rs
```diff
@@ -236,8 +240,8 @@ fn log_negative_user_reaction(
     embedding: &Embedding,
     config: &Configuration,
 ) {
-    match find_closest_coi_mut(cois, embedding, config.neighbors()) {
-        Some((coi, distance)) if distance < config.threshold() => {
+    match find_closest_coi_mut(cois, embedding) {
+        Some((coi, distance)) if distance >= config.threshold() => {
```
janpetschexain, line 244: it's still called `distance` instead of `similarity`
```suggestion
        Some((coi, similarity)) if similarity >= config.threshold() => {
```

xayn-ai/src/coi/system.rs
```diff
@@ -289,10 +293,10 @@ pub(crate) fn update_user_interests(
 fn log_document_view_time(
     cois: &mut Vec<PositiveCoi>,
     embedding: &Embedding,
-    config: &Configuration,
+    _config: &Configuration,
```
janpetschexain, line 296: if we don't need the config anymore here, we could just remove the argument.

xayn-ai/src/embedding/utils.rs
```diff
@@ -53,9 +54,17 @@ where
     mean
 }
 
+/// See [`pairwise_cosine_similarity`] for details
+pub const MAXIMUM_COSINE_SIMILARITY: f32 = 1.0;
+pub const MINIMUM_COSINE_SIMILARITY: f32 = -1.0;
+pub const COSINE_SIMILARITY_RANGE: RangeInclusive<f32> =
+    MINIMUM_COSINE_SIMILARITY..=MAXIMUM_COSINE_SIMILARITY;
```
janpetschexain, line 61: it's a good idea to state them as constants. now we can also use them in the clamping in l.99.

xayn-ai/src/embedding/utils.rs
```diff
@@ -97,15 +105,9 @@ where
 
 /// Computes the cosine similarity of two vectors.
 ///
-/// Zero vectors are chosen to be similar to all other vectors, i.e. a similarity of 1.
-///
-/// # Panics
-/// Panics if the vectors don't consist solely of real values or their shapes don't match.
+/// See [`pairwise_cosine_similarity`] for details.
 #[allow(dead_code)]
```
janpetschexain, line 109: the attribute can still be removed
```suggestion
```

janpetschexain (commented):
lgtm, some small leftovers

xayn-ai/src/coi/system.rs
x3ro: I wasn't sure whether we'd introduce other configuration that will be relevant here, and since it's a cascading change I just left it in for now. Do you think this is a must-have for getting this merged?

janpetschexain: not a must have, we can change it later on if needed as well.

x3ro:
bors merge

janpetschexain (approved):

x3ro:
bors merge

PR #389, opened 2022-01-20
[Merged by Bors] - FIX: example apps

janpetschexain:
**References**

- unblocks #396

**Summary**

- add missing field to ffi & update example call data


bindings/dart/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -363,6 +363,7 @@
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
+				ONLY_ACTIVE_ARCH = YES;
```
Robert-Steiner, line 366: was this automatically added by xcode?
because normally that is only set for debug but not for profile or release build.

janpetschexain: i think this was added by my xcode when i tried to make it work that xcode builds&runs the flutter app (via the gui). but since this still doesn't work and we have to call `flutter run` anyways, i'll try to revert this.

bindings/dart/ffigen_common.yaml
```diff
@@ -6,6 +6,3 @@ headers:
     - 'ios/Classes/XaynAiFfiCommon.h'
   include-directives:
     - 'ios/Classes/XaynAiFfiCommon.h'
-compiler-opts-automatic:
```
acrrd, line 9: Is this ok now? @rustonaut We may need to change this in the discovery engine

Robert-Steiner (approved):
From my side it looks good

bindings/dart/ffigen_common.yaml
janpetschexain: compilation fails if set to `false` now, so i had to set it to `true`. the generated output seems to be the same. i'm not sure if this was needed before (maybe to mirror the settings in `cbindgen`?) and i also don't know why it worked before and not anymore.

janpetschexain:
bors merge

PR #399, opened 2022-01-24
[Merged by Bors] - TY-2303 change relevance maps [1]

janpetschexain:
**References**

- [TY-2303]
- #398

**Summary**

- add type safety for relevance variants: the new `Relevance`/`s` wrappers internally maintain two variants for coi and key phrase relevances respectively
- some renaming: `Relevance -> F32`, `Relevances -> RelevanceMap`

[TY-2303]: https://xainag.atlassian.net/browse/TY-2303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/xayn-ai/src/coi/relevance.rs
+++ b/xayn-ai/src/coi/relevance.rs
@@ -33,9 +33,12 @@ impl Ord for F32 {
 #[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
 pub(crate) struct Relevance(Rel);
 
+/// Relevance score variants.
 #[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
 enum Rel {
+    /// A coi relevance score.
     Coi(F32),
+    /// A key phrase relevance score.
     Kp(F32),
 }
 
@@ -78,9 +81,12 @@ impl From<Relevance> for f32 {
 #[derive(Debug)]
 pub(crate) struct Relevances(Rels);
 
+/// Relevance scores variants.
 #[derive(Debug)]
 enum Rels {
+    /// Coi relevance scores.
     Coi(F32),
+    /// Key phrases relevance scores.
     Kps(BTreeSet<F32>),
 }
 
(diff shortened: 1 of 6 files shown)


xayn-ai/src/coi/relevance.rs
```diff
@@ -7,49 +7,91 @@ use derive_more::Into;
...
+}
+
+/// A relevance score.
+#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
+pub(crate) struct Relevance(Rel);
+
+#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
+enum Rel {
+    Coi(F32),
+    Kp(F32),
```
Robert-Steiner, line 39: Maybe we can add a doc string here that kp stands for key phrase

xayn-ai/src/coi/relevance.rs
```diff
@@ -7,49 +7,91 @@ use derive_more::Into;
...
+        self.partial_cmp(other).unwrap(/* never nan by invariant */)
+    }
+}
+
+/// A relevance score.
+#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
+pub(crate) struct Relevance(Rel);
+
+#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
+enum Rel {
```
Robert-Steiner, line 37: What about impl the `coi` and `kp` methods on the enum and removing the `Relevance` struct?

Robert-Steiner (approved):
Looks good just a question if we can remove the `Relevance` and `Relevances` structs.

xayn-ai/src/coi/relevance.rs
janpetschexain: in this case the issue is with enum visibility. since we would need to make it `pub(crate) enum Relevance` which means that the variants and especially the contained data is a public field too. in turn we would have to make `F32` public as well, otherwise we get the private-type-in-public-interface compiler error, but this is an internal implementational detail (in the end it's the whole purpose of the enum wrapping of coi/key phrase relevances that outside of this module noone has to deal with the internals, but if we were to expose it then people could match against the enum and split the fields apart out of context rendering the additional type safety void).

it would be nice to have public enums where variants can contain private associated data. i tried to work around that by wrapping a private enum in a public struct.

janpetschexain:
bors merge

PR #425, opened 2022-02-22
[Merged by Bors] - TY 2372 Sort when not enough coi

acrrd:
If we don't have enough coi we use the score or the rank to sort.
If both document have the score we use the score, otherwise we use the rank.
Higher score is better. Lower rank is better.


x3ro (approved):

Robert-Steiner (approved):

acrrd:
bors merge

PR #431, opened 2022-02-28
[Merged by Bors] - Insert cleaned key phrases when map is empty

acrrd:
When we don't have any key phrase to return we insert again all cleaned one.

### DIFF
--- a/xayn-ai/src/coi/key_phrase.rs
+++ b/xayn-ai/src/coi/key_phrase.rs
@@ -110,6 +110,8 @@ impl RelevanceMap {
         horizon: Duration,
         penalty: &[f32],
     ) -> Vec<KeyPhrase> {
+        self.insert_cleaned_if_empty();
+
         self.compute_relevances(cois, horizon, system_time_now());
 
         // TODO: refactor once pop_last() etc are stabilized for BTreeMap
--- a/xayn-ai/src/coi/key_phrase.rs
+++ b/xayn-ai/src/coi/key_phrase.rs
@@ -123,7 +123,8 @@ impl RelevanceMap {
                     (coi_id, relevance, penalized_relevance, key_phrase.clone())
                 })
             }).collect::<Vec<_>>();
-        relevances.sort_unstable_by(|(_, _, this, _), (_, _, other, _)| this.cmp(other).reverse());
+        relevances.sort_by(|(_, _, this, _), (_, _, other, _)| this.cmp(other).reverse());
+
         relevances
             .into_iter()
             .take(top)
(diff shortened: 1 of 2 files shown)


xayn-ai/src/coi/relevance.rs
```diff
@@ -144,6 +145,7 @@ impl RelevanceMap {
         self.coi_to_relevance
             .remove(&coi_id)
             .map(|relevances| {
+                self.cleaned.remove(&coi_id);
```
janpetschexain, line 148: is there a specific reason that this call is inside the closure? otherwise it could be called outside.

xayn-ai/src/coi/key_phrase.rs
```diff
@@ -121,7 +123,8 @@ impl RelevanceMap {
                     (coi_id, relevance, penalized_relevance, key_phrase.clone())
                 })
             }).collect::<Vec<_>>();
-        relevances.sort_unstable_by(|(_, _, this, _), (_, _, other, _)| this.cmp(other).reverse());
+        relevances.sort_by(|(_, _, this, _), (_, _, other, _)| this.cmp(other).reverse());
```
acrrd, line 126: To have consistency in tests. Does not really affect performance for us.

xayn-ai/src/coi/relevance.rs
janpetschexain, line 148: removing the id from the `cleaned` might starve this map soon. the main issue is that `RelevanceMap::remove()` is called in every iteration of the feedback loop (via `RelevanceMap::select_key_phrases()`), but while the key phrases of the corresponding id are removed & unified with new candidates & selected & inserted again in the `coi_to_relevance`/`relevance_to_key_phrase` maps, this is not the case yet for the `cleaned` map which will starve it. on the other hand, if we ran another iteration of the feedback loop, than we have new key phrases and don't have to worry about the `cleaned` anymore, is this the idea?

acrrd: In the feedback loop we are removing the coi and put it back with updated data, so the coi is not clean anymore but we have fresh data in and the keywords that we just put in can be selected. 
After the remove the information about that coi in `cleaned` are stale.

xayn-ai/src/coi/relevance.rs
```diff
@@ -169,6 +171,10 @@ impl RelevanceMap {
     pub(super) fn clean(&mut self, coi_id: CoiId, relevance: Relevance, key_phrase: &KeyPhrase) {
         if let Some(key_phrases) = self.relevance_to_key_phrase.get_mut(&(relevance, coi_id)) {
             key_phrases.retain(|this| this != key_phrase);
+            self.cleaned
+                .entry(coi_id)
+                .or_default()
+                .push((relevance, key_phrase.clone()));
```
janpetschexain, line 177: it seems wasteful to clone the key phrase which we have just removed. instead of `key_phrases.retain()` we could find the index of the `key_phrase` in the `key_phrases` via `Iterator::position()` and `Vec::remove()` it and then `push()` it to the `cleaned` map.

this will also solve the remaining issue that even though the pushing is now inside the `if let`, it is still possible that we push a key phrase to `cleaned` which never existed in `key_phrases` before (the retaining will not remove anything if the `key_phrase` doesn't exist).

overall, we probably need more tests for the `RelevanceMap` methods to cover all those edge cases, maybe in a follow up pr.

xayn-ai/src/coi/relevance.rs
```diff
@@ -251,6 +257,20 @@ impl RelevanceMap {
...
         }
     }
+
+    pub(super) fn insert_cleaned_if_empty(&mut self) {
+        if self.relevance_to_key_phrase.is_empty() {
+            let to_insert = std::mem::take(&mut self.cleaned);
+
+            for (coi_id, value) in to_insert.into_iter() {
+                // we need to insert in reverse order to keep the order in which
+                // the key phrases were originally
```
janpetschexain, line 267: i double checked this, we could update the comment and point to the exact reason why we need to reverse. in `RelevanceMap::select_top_key_phrases()` we collect and sort the key phrases in descending order wrt the penalized relevances, then the selected key phrases are cleaned in that reversed order.

it's also important that this simple reverse only works if we keep up the invariant that the `penalty` must be sorted decreasingly. if we ever need to change that, then `cleaned` needs to also track the penalized relevances and reverse the insert order wrt them.

janpetschexain (approved):
👍

acrrd:
bors merge

PR #442, opened 2022-03-10
TY-2467 add cosine distance

Robert-Steiner:
Ticket:

- [TY-2467]


xayn-ai/src/embedding/utils.rs
```diff
@@ -110,6 +110,11 @@ pub fn cosine_similarity(a: ArrayView1<f32>, b: ArrayView1<f32>) -> f32 {
     pairwise_cosine_similarity([a.view(), b.view()])[[0, 1]]
 }
 
+/// Computes the cosine distance of two vectors.
+pub fn cosine_distance(a: ArrayView1<f32>, b: ArrayView1<f32>) -> f32 {
```
acrrd, line 114: Can't we have this in the de instead of splitting the code?

Robert-Steiner: yes we can do this. the only downside is that we have to add `ndarray` dependency back to de

acrrd: I don't think is a problem

x3ro (approved):

